PR #9, opened 2018-10-30
Accordian Spike

ed0906:
A Spike to create an accordian template. Based on an extension of the input_radio twirl template.

### DIFF
--- a/app/assets/javascripts/show-hide-content.js
+++ b/app/assets/javascripts/show-hide-content.js
@@ -13,7 +13,7 @@
     // Radio and Checkbox selectors
     var selectors = {
       namespace: 'ShowHideContent',
-      radio: '.block-label[data-target] input[type="radio"]',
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/show-hide-content.js
```diff
@@ -13,7 +13,7 @@
     // Radio and Checkbox selectors
     var selectors = {
       namespace: 'ShowHideContent',
-      radio: '.block-label[data-target] input[type="radio"]',
+      radio: '[data-target] > input[type="radio"]',
```
ed0906: For whatever reason other front end apps seem to have this ^^ in the `show-hide-content.js` is `block-label` a legacy class?

gokyo: I do not know.
You could ask in the `#community-frontend` Slack channel

PR #10, opened 2018-10-31
Read Me for Service Manager

ed0906:


README.md
```diff
@@ -13,13 +13,15 @@ The Front End for the BTI Application journey on GOV.UK
 
 ##### Starting the application:
 
-1) Run [Binding Tariff Classification](https://github.com/hmrc/binding-tariff-classification) on Port 9090: `sbt 'run 9090'`
-2) Start Assets Frontend Using `sm --start ASSETS_FRONTEND -r 3.2.2`
+1) Start Assets Frontend Using `sm --start ASSETS_FRONTEND -r 3.2.2`
```
gokyo: I believe we want to use the latest version of ASSETS_FRONTEND (something like 4.5.0)
Am I right?

README.md
```diff
@@ -13,13 +13,15 @@ The Front End for the BTI Application journey on GOV.UK
...
+2) Start [Binding Tariff Classification](https://github.com/hmrc/binding-tariff-classification) Using `sm --start BINDING_TARIFF_CLASSIFICATION -f`
 3) Start Auth Using `sm --start AUTH -f`
 4) Start Auth Login Stub `sm --start AUTH_LOGIN_STUB -f`
 5) Start Auth Login API `sm --start AUTH_LOGIN_API -f`
 6) Start User Details `sm --start USER_DETAILS -f`
 
+Finally Run `sbt run` to boot the app
```
gokyo: And if I want to run it with `sm`?

README.md
```diff
@@ -13,13 +13,15 @@ The Front End for the BTI Application journey on GOV.UK
...
+
 Go to the [Start Page](http://localhost:9000/binding-tariff-trader-frontend/registeredAddressForEori).
 You will be redirected to the Auth Stub. Make sure you pick `Affinity=Organisation` and Submit.
```
gokyo: Add instructions for running the tests

README.md
ed0906: The front end requires that specific version according to @russellthorn

gokyo: ok

russellthorn: Scaffold comes pre-packaged using version 3.2.2, I don't think we're tied to that. Bumping to version 4 may have a knock on effect on styling and may require changes to class names to adopt the same styling due to differences in v3 VS v4. I haven't used v4 much. I did try implement v4 in the early build of imports, but ended up reverting the the latest version of v3 because had some issues with v4 of AF.

In all honesty, There's little v4 gives us over v3, possibly nothing. I don't know off the top of my head. I would suggest if anything, bumping to the latest version of v3 for now. Which has zero risk of requiring any further work and only bumping versions away from v3 when v5 (will be known as HMRC frontend) is released. As I understand it the new version is currently being produced and will be inline and keeping up with the latest assets frontend from GOV.UK, simply called Frontend. Which the new design system is built with. Mainly new class names etc using BEM naming conventions.

PR #21, opened 2018-11-28
DIT-487: explicit audit events

gokyo:


app/controllers/DeclarationController.scala
```diff
@@ -49,15 +52,35 @@ class DeclarationController @Inject()(
...
+      .map(Some(_))
+      .recover(handleFailingSubmission(newCaseRequest))
+      .flatMap {
+        case Some(c: Case) =>
+          auditService.auditBTIApplicationSubmissionSuccessful(c)
+          userAnswers.set(ConfirmationPage, Confirmation(c.reference))
+
+          dataCacheConnector.save(userAnswers.cacheMap) map ( _ => redirect(mode, userAnswers) )
+        case _ =>
+          Future.successful(redirect(mode, userAnswers))
```
ed0906: If the case create fails do we actually want to redirect to the confirmation page?
I would have thought it would be better to go to an error page, or back to the declaration page so they can try again

PR #29, opened 2018-12-11
DIT-72 Require CDS enrolment when you log in

pllaurado:


app/controllers/actions/IdentifierAction.scala
```diff
@@ -33,11 +33,14 @@ import scala.concurrent.{ExecutionContext, Future}
...
 
+
+  lazy val requiredEnrolment = Enrolment(config.authEnrolment)
+
   override def invokeBlock[A](request: Request[A], block: IdentifierRequest[A] => Future[Result]): Future[Result] = {
 
     implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromHeadersAndSession(request.headers, Some(request.session))
 
-    authorised().retrieve(Retrievals.internalId) {
+    authorised(requiredEnrolment).retrieve(Retrievals.internalId) {
```
gokyo: Can you check also it comes from GG?
Should be:
```
authorised(Enrolment("HMRC-CUS-ORG") and AuthProviders(GovernmentGateway))
```

See this example:
https://github.com/hmrc/customs-inventory-linking-exports/blob/1d6bf8d2facf1689ce73653e5d168ac318268ad1/app/uk/gov/hmrc/customs/inventorylinking/export/controllers/actionbuilders/AuthAction.scala#L130-L136

pllaurado: It's a right assumption, we need to use the Government Gateway  but this come as default in Authorization. This is why right now auth is working without any param.  You could check it on the unit test 

`    "the user hasn't logged in" must {
      "redirect the user to log in " in {
        val authAction = new AuthenticatedIdentifierAction(new FakeFailingAuthConnector(new MissingBearerToken), frontendAppConfig)
        val controller = new Harness(authAction)
        val result = controller.onPageLoad()(fakeRequest)
        status(result) mustBe SEE_OTHER
        redirectLocation(result).get must startWith(frontendAppConfig.loginUrl)
      }
    }

    "the user's session has expired" must {
      "redirect the user to log in " in {
        val authAction = new AuthenticatedIdentifierAction(new FakeFailingAuthConnector(new BearerTokenExpired), frontendAppConfig)
        val controller = new Harness(authAction)
        val result = controller.onPageLoad()(fakeRequest)
        status(result) mustBe SEE_OTHER
        redirectLocation(result).get must startWith(frontendAppConfig.loginUrl)
      }
    }`

gokyo:
Are you sure there are enough tests for this feature?

PR #30, opened 2018-12-13
BAU improve Logging and comments around un-scanned attachments

ed0906:


test/unit/service/FileServiceSpec.scala
```diff
@@ -93,4 +93,37 @@ class FileServiceSpec extends UnitSpec with MockitoSugar with BeforeAndAfterEach
       await(service.publish(filePublishing)) shouldBe UnpublishedFileAttachment("id", "filename-updated", "type", 0, "Quarantined")
     }
   }
+
+  "Publish multiple" should {
```
gokyo: I would change the tests.
I would have the first item of the sequence failing and the second working (rather than the other way around - what you wrote now)

ed0906: Done

PR #39, opened 2019-01-08
DIT-659 Modify Creating Attachments

ed0906:


app/service/FileService.scala
```diff
@@ -40,48 +40,27 @@ class FileService @Inject()(connector: BindingTariffFilestoreConnector) {
...
       files.map { f: FileAttachment =>
-        publish(f).recover {
-          case t: Throwable => UnpublishedFileAttachment(f.id, f.name, f.mimeType, f.size, t.getMessage)
+        publish(f).map(Some(_)).recover {
+          case t: Throwable => Logger.error(s"Failed to publish file [${f.id}].", t)
+            None
         }
       }
-    )
+    ).map(_.filter(_.isDefined).map(_.get))
```
gokyo: Seems unnecessary to wrap the files in an `Option` and then call `.get`
There must be a better way to do it. 
One idea I have in my mind right now is to have an extra field of type `Option[Boolean]` with default value `false` in `PublishedFileAttachment`. When the file is published, you update that field.

ed0906: It's because of the recover block

ed0906: I'd stick the recover at the end but then it would fail all the attachments if one of them failed

PR #47, opened 2019-01-29
DIT-710: auditing only necessary fields (no status, no attachments)

gokyo:


app/models/Case.scala
```diff
@@ -28,3 +28,9 @@ case class Case
   application: Application,
   attachments: Seq[Attachment] = Seq.empty
 )
+
+case class CaseAuditPayload
```
ed0906: Could this not be in a different file rather than sticking it all in the Case model file

gokyo: done

PR #48, opened 2019-02-06
Proposed fix for validating file type of agent authorisation letter u…

SteveSugden:
…pload.

### DIFF
--- a/app/config/FrontendAppConfig.scala
+++ b/app/config/FrontendAppConfig.scala
@@ -51,6 +51,9 @@ class FrontendAppConfig @Inject() (override val runModeConfiguration: Configurat
   lazy val bindingTariffFileStoreUrl: String = baseUrl("binding-tariff-filestore")
   lazy val authEnrolment: Option[String] = Some(loadConfig("auth.enrolment")).filter(isNoneBlank(_))
 
+  lazy val fileUploadMaxSize = loadConfig("fileupload.maxSize").toInt
+  lazy val fileUploadMimeTypes = loadConfig("fileupload.mimeTypes")
+
   lazy val languageTranslationEnabled = runModeConfiguration.getBoolean("microservice.services.features.welsh-translation").getOrElse(true)
 
   def languageMap: Map[String, Lang] = Map(
--- a/app/config/FrontendAppConfig.scala
+++ b/app/config/FrontendAppConfig.scala
@@ -52,7 +52,7 @@ class FrontendAppConfig @Inject() (override val runModeConfiguration: Configurat
   lazy val authEnrolment: Option[String] = Some(loadConfig("auth.enrolment")).filter(isNoneBlank(_))
 
   lazy val fileUploadMaxSize = loadConfig("fileupload.maxSize").toInt
-  lazy val fileUploadMimeTypes = loadConfig("fileupload.mimeTypes")
+  lazy val fileUploadMimeTypes = loadConfig("fileupload.mimeTypes").split(",").map(_.trim).toSet
 
   lazy val languageTranslationEnabled = runModeConfiguration.getBoolean("microservice.services.features.welsh-translation").getOrElse(true)
 
--- a/app/controllers/UploadWrittenAuthorisationController.scala
+++ b/app/controllers/UploadWrittenAuthorisationController.scala
@@ -95,12 +95,37 @@ class UploadWrittenAuthorisationController @Inject()(
     }
 
   private val maxFileSize = 10485760
+  private val allowedTypes = Set(
+    "application/pdf",
+    "application/msword",
+    "application/vnd.ms-excel",
+    "image/png",
+    "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
+    "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+    "image/jpeg",
+    "text/plain"
+  )
+
   private def valid(file: MultipartFormData.FilePart[TemporaryFile]): Either[String, MultipartFormData.FilePart[TemporaryFile]] = {
-    if (file.ref.file.length() > maxFileSize) {
-      Left(messagesApi("uploadWrittenAuthorisation.error.size"))
-    } else {
-      Right(file)
+
+    val error = validFileSize(file).orElse(validFileType(file))
+    error.fold(Right(file): Either[String, MultipartFormData.FilePart[TemporaryFile]]) {
+      Left(_)
     }
   }
 
+  private def validFileSize(file: MultipartFormData.FilePart[TemporaryFile]): Option[String] = {
+    if (file.ref.file.length() > maxFileSize)
+      Some(messagesApi("uploadWrittenAuthorisation.error.size"))
+    else
+      None
+  }
+
+  private def validFileType(file: MultipartFormData.FilePart[TemporaryFile]): Option[String] = {
+
+    if (file.contentType.exists(t => allowedTypes.contains(t)))
+      None
+    else
+      Some(messagesApi("uploadWrittenAuthorisation.error.fileType"))
+  }
 }
--- a/app/controllers/UploadWrittenAuthorisationController.scala
+++ b/app/controllers/UploadWrittenAuthorisationController.scala
@@ -82,7 +82,7 @@ class UploadWrittenAuthorisationController @Inject()(
 
       letterOfAuthority match {
         case Some(file) =>
-          valid(file) match {
+          fileService.validate(file) match {
             case Right(validFile) => uploadFile(validFile)
             case Left(errorMessage) => badRequest("validation-error", errorMessage)
           }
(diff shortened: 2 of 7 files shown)


app/controllers/UploadWrittenAuthorisationController.scala
```diff
@@ -95,12 +95,37 @@ class UploadWrittenAuthorisationController @Inject()(
...
+  private def validFileSize(file: MultipartFormData.FilePart[TemporaryFile]): Option[String] = {
+    if (file.ref.file.length() > maxFileSize)
+      Some(messagesApi("uploadWrittenAuthorisation.error.size"))
+    else
+      None
+  }
+
+  private def validFileType(file: MultipartFormData.FilePart[TemporaryFile]): Option[String] = {
+
+    if (file.contentType.exists(t => allowedTypes.contains(t)))
```
pllaurado: I would do it in the other way allowedTypes.contains(file.contentType)

app/controllers/UploadWrittenAuthorisationController.scala
```diff
@@ -95,12 +95,37 @@ class UploadWrittenAuthorisationController @Inject()(
     }
 
   private val maxFileSize = 10485760
+  private val allowedTypes = Set(
```
gokyo: We need to add the same validation in the operational UI I believe....
Is it worth doing this maybe in `binding-tariff-filestore`, so that is used by both front-end microservices?

app/controllers/UploadWrittenAuthorisationController.scala
```diff
@@ -95,12 +95,37 @@ class UploadWrittenAuthorisationController @Inject()(
...
-    } else {
-      Right(file)
+
+    val error = validFileSize(file).orElse(validFileType(file))
+    error.fold(Right(file): Either[String, MultipartFormData.FilePart[TemporaryFile]]) {
+      Left(_)
     }
   }
 
+  private def validFileSize(file: MultipartFormData.FilePart[TemporaryFile]): Option[String] = {
```
pllaurado: Maybe this would be easier. If we want a more scalable solution other approaches would work better, but I don't think we will need more validations.
` if (file.ref.file.length() > maxFileSize)
      Left ( Max size error Message)
  if (allowedTypes.contains(file.contentType))
     Left  ( Wrong file type Message)

  Right (file)
`

app/controllers/UploadWrittenAuthorisationController.scala
pllaurado: We need in both, but this does not make filestore the place where to validate. We could make a third service which could allow any type of file for example.

app/service/FileService.scala
```diff
@@ -55,6 +56,19 @@ class FileService @Inject()(connector: BindingTariffFilestoreConnector) {
...
+  def validate(file: MultipartFormData.FilePart[TemporaryFile]): Either[String, MultipartFormData.FilePart[TemporaryFile]] = {
+
+    file match {
+      case f if hasWrongSize (f) => Left(messagesApi("uploadWrittenAuthorisation.error.size"))
+      case f if hasWrongContentType (f) => Left(messagesApi("uploadWrittenAuthorisation.error.fileType"))
+      case _ => Right(file)
+    }
+
+  }
+  private def hasWrongSize : MultipartFormData.FilePart[TemporaryFile] => Boolean = { file  => file.ref.file.length() > maxFileSize}
```
gokyo: can be improved as 
```
{ _.ref.file.length > maxFileSize }
```

app/service/FileService.scala
```diff
@@ -55,6 +56,19 @@ class FileService @Inject()(connector: BindingTariffFilestoreConnector) {
...
+  private def hasWrongContentType : MultipartFormData.FilePart[TemporaryFile] => Boolean = {file  => file.contentType.filter(allowedTypes.contains).isEmpty}
```
gokyo: can be improved as 
```
{ _.contentType.filter(allowedTypes.contains).isEmpty }
```

But I would say:
```
!allowedTypes.contains(_.contentType)
```

test/unit/controllers/UploadWrittenAuthorisationControllerSpec.scala
```diff
@@ -136,6 +137,20 @@ class UploadWrittenAuthorisationControllerSpec extends ControllerSpecBase with M
...
+      val file = TemporaryFile("example-file.mp3")
+      val filePart = FilePart[TemporaryFile](key = "letter-of-authority", "example-file.mp3", contentType = Some("audio/mpeg"), ref = file)
+      val form = MultipartFormData[TemporaryFile](dataParts = Map(), files = Seq(filePart), badParts = Seq.empty)
+      val postRequest = fakeRequest.withBody(form)
+
+      given(fileService.validate(any[MultipartFormData.FilePart[TemporaryFile]])).willReturn(Left("bad file"))
+
+      val result = controller().onSubmit(NormalMode)(postRequest)
+
+      status(result) mustBe BAD_REQUEST
```
gokyo: how about the body of the response?

app/service/FileService.scala
SteveSugden: ContentType is an Option[String] so the allowedTypes won't contain it.  But I'll refactor both function definitions as you suggest

app/service/FileService.scala
```diff
@@ -63,4 +77,15 @@ class FileService @Inject()(connector: BindingTariffFilestoreConnector) {
     r => PublishedFileAttachment(r.id, r.fileName, r.mimeType, size)
   }
 
+  private val maxFileSize = 10485760
+  private val allowedTypes = Set(
```
gokyo: I would define these extensions in the config file (`application.conf`) and then read it in memory using `AppConfig.scala`.
This would allow us to have different allowed types for each environment (local, QA, production).
Also, if in the future we want to remove or add a file type, we would just need to change the config

SteveSugden: That is a good idea.  Will change

app/service/FileService.scala
```diff
@@ -63,4 +78,5 @@ class FileService @Inject()(connector: BindingTariffFilestoreConnector) {
     r => PublishedFileAttachment(r.id, r.fileName, r.mimeType, size)
   }
 
+  private lazy val allowedTypes = configuration.fileUploadMimeTypes.split(",").map(_.trim).toSet
```
ed0906: This split logic should probably exist in the `FrontendConfig` class

test/unit/controllers/UploadWrittenAuthorisationControllerSpec.scala
ed0906: Looks like he is below @gokyo

app/config/FrontendAppConfig.scala
```diff
@@ -51,6 +51,9 @@ class FrontendAppConfig @Inject() (override val runModeConfiguration: Configurat
   lazy val bindingTariffFileStoreUrl: String = baseUrl("binding-tariff-filestore")
   lazy val authEnrolment: Option[String] = Some(loadConfig("auth.enrolment")).filter(isNoneBlank(_))
 
+  lazy val fileUploadMaxSize = loadConfig("fileupload.maxSize").toInt
+  lazy val fileUploadMimeTypes = loadConfig("fileupload.mimeTypes")
```
ed0906: I think the split logic should be done here, rather than when it is used

test/unit/controllers/UploadWrittenAuthorisationControllerSpec.scala
SteveSugden: I added that assertion after Stefano's comment.

test/unit/controllers/UploadWrittenAuthorisationControllerSpec.scala
```diff
@@ -136,6 +138,22 @@ class UploadWrittenAuthorisationControllerSpec extends ControllerSpecBase with M
...
+    "return a Bad Request when invalid file type is submitted" in {
+
+      val file = TemporaryFile("example-file.mp3")
+      val filePart = FilePart[TemporaryFile](key = "letter-of-authority", "example-file.mp3", contentType = Some("audio/mpeg"), ref = file)
+      val form = MultipartFormData[TemporaryFile](dataParts = Map(), files = Seq(filePart), badParts = Seq.empty)
+      val postRequest = fakeRequest.withBody(form)
+
+      given(fileService.validate(any[MultipartFormData.FilePart[TemporaryFile]])).willReturn(Left("some error message about bad file"))
+
+      val result = controller().onSubmit(NormalMode)(postRequest)
```
gokyo: we typically execute the requests with `await` - but I close an eye here

app/config/FrontendAppConfig.scala
```diff
@@ -51,6 +51,9 @@ class FrontendAppConfig @Inject() (override val runModeConfiguration: Configurat
   lazy val bindingTariffFileStoreUrl: String = baseUrl("binding-tariff-filestore")
   lazy val authEnrolment: Option[String] = Some(loadConfig("auth.enrolment")).filter(isNoneBlank(_))
 
+  lazy val fileUploadMaxSize = loadConfig("fileupload.maxSize").toInt
```
gokyo: we should test this, but I close an eye this time

app/service/FileService.scala
```diff
@@ -47,14 +49,32 @@ class FileService @Inject()(connector: BindingTariffFilestoreConnector) {
...
-        publish(f).map(Some(_)).recover {
-          case t: Throwable => Logger.error(s"Failed to publish file [${f.id}].", t)
+        publish(f).map(Option(_)).recover {
+          case t: Throwable =>
+            Logger.error(s"Failed to publish file [${f.id}].", t)
             None
         }
       }
     ).map(_.filter(_.isDefined).map(_.get))
   }
```
gokyo: I did some cleaning here

PR #50, opened 2019-02-12
Dit 715 ui changes

SteveSugden:


SteveSugden:
This is the PR to fix the acceptance tests -0 https://github.com/hmrc/digital-tariffs-acceptance-tests/pull/106

app/views/components/input_textarea.scala.html
```diff
@@ -29,6 +30,9 @@
         @hint.map { hint =>
             <span class="form-hint">@hint</span>
         }
+        @htmlHint.map { htmlHint =>
+            <span class="form-hint">@htmlHint</span>
```
ed0906: Minor, but is a span the right html element here? Looks more like a `div` scenario to me

app/views/confirmation.scala.html
```diff
@@ -23,34 +22,39 @@
...
     bodyClasses = None) {
 
-	  <div class="govuk-box-highlight">
-			@components.heading("confirmation.heading", "heading-xlarge")
-      <p class="font-large">
-    		@messages("confirmation.yourRefNumber")<br>
-        <span id="confirmation-reference" class="bold">@messages("confirmation.refValue", confirmation.reference)</span>
-      </p>
-	  </div>
+    <div class="govuk-box-highlight">
```
ed0906: Does this class do anything?
Generally we don't have access to `govuk-` classes from the design system

app/views/components/input_textarea.scala.html
SteveSugden: You have a good point.  I'll change

PR #60, opened 2019-02-22
DIT-789: retrieving EORI number

gokyo:


app/controllers/actions/IdentifierAction.scala
```diff
@@ -43,10 +46,18 @@ class AuthenticatedIdentifierAction @Inject()(override val authConnector: AuthCo
...
-    authorise().retrieve(Retrievals.internalId) {
-      _.map {
-        internalId => block(IdentifierRequest(request, internalId))
-      }.getOrElse(throw new UnauthorizedException("Unable to retrieve internal Id"))
+    // TODO: remove fake EORI number when in public beta
+    def eoriNumber(enrolments: Enrolments): String = {
+      enrolments.getEnrolment(cdsEnrolment).flatMap(_.getIdentifier("EORINumber")) match {
+        case Some(eori) => eori.value
+        case _ if config.isCdsEnrolmentCheckEnabled => throw InsufficientEnrolments()
+        case _ => "GB6723190" // fake EORI number
```
SteveSugden: why do we need this fake hard codes value now?

gokyo: Because in the private beta phase we want to allow users without a EORI number to use the application

app/views/confirmation.scala.html
```diff
@@ -24,7 +24,7 @@
     bodyClasses = None) {
 
     <div class="govuk-box-highlight">
-        @components.heading("confirmation.heading", "heading-xlarge")
+        @components.heading("confirmation.heading")
```
gokyo: `heading-xlarge` is the default value in `heading.scala` - so need to specify it here

app/views/registeredAddressForEori.scala.html
```diff
@@ -31,7 +32,9 @@
 
         @components.error_summary(form.errors)
 
-        @components.heading("registeredAddressForEori.heading")
+        @components.heading(
```
gokyo: for some reason the EORI number is not shown here... Any tip to fix this?

test/unit/views/RegisteredAddressForEoriViewSpec.scala
```diff
@@ -46,6 +49,18 @@ class RegisteredAddressForEoriViewSpec extends QuestionViewBehaviours[Registered
       routes.RegisteredAddressForEoriController.onSubmit(NormalMode).url,
       "field1", "field2", "field3", "field4", "field5"
     )
+
+    "show the expected text" in {
+      val text = asDocument(createViewUsingForm(form)).text()
+
+      text must include("What is the registered name and address for EORI number eori-789012?")
```
gokyo: this is failing currently
It shows `What is the registered name and address for EORI number {0}`

Any idea on how to fix it?

conf/messages.en
```diff
@@ -62,7 +62,6 @@ site.continue = Continue
 site.submit = Submit
 site.upload_continue = Upload and continue
 site.service_name = Apply for a Binding Tariff Information ruling
-site.textarea.char_limit = (Limit is {0} characters)
```
gokyo: this is not used

app/views/registeredAddressForEori.scala.html
gokyo: I think the problem is that `heading.scala` expects a messages key, not a messages value.
Can we make `heading.scala` more generic so that it takes also a sequence of optional String arguments to pass when calling `messages`?

gokyo: I tried making `heading.scala` more generic, but I can't make the test passing...

app/controllers/actions/DataRetrievalAction.scala
```diff
@@ -33,9 +34,8 @@ class DataRetrievalActionImpl @Inject()(val dataCacheConnector: DataCacheConnect
 
     implicit val hc: HeaderCarrier = HeaderCarrierConverter.fromHeadersAndSession(request.headers, Some(request.session))
 
-    dataCacheConnector.fetch(request.identifier).map {
-      case None => OptionalDataRequest(request.request, request.identifier, None)
```
gokyo: avoiding repetitions

app/config/FrontendAppConfig.scala
```diff
@@ -49,7 +48,7 @@ class FrontendAppConfig @Inject() (override val runModeConfiguration: Configurat
   lazy val loginContinueUrl: String = loadConfig("urls.loginContinue")
   lazy val bindingTariffClassificationUrl: String = baseUrl("binding-tariff-classification")
   lazy val bindingTariffFileStoreUrl: String = baseUrl("binding-tariff-filestore")
-  lazy val authEnrolment: Option[String] = Some(loadConfig("auth.enrolment")).filter(isNoneBlank(_))
+  lazy val isCdsEnrolmentCheckEnabled: Boolean = getBoolean("cdsEnrolmentCheckEnabled")
```
gokyo: a boolean config is simpler

app/config/Module.scala
```diff
@@ -27,10 +27,7 @@ class Module extends AbstractModule {
     // Bind the actions for DI
     bind(classOf[DataRetrievalAction]).to(classOf[DataRetrievalActionImpl]).asEagerSingleton()
     bind(classOf[DataRequiredAction]).to(classOf[DataRequiredActionImpl]).asEagerSingleton()
-
-    // For session based storage instead of cred based, change to SessionIdentifierAction
```
gokyo: I have deleted `SessionIdentifierAction` now

app/controllers/actions/IdentifierAction.scala
```diff
@@ -58,28 +69,11 @@ class AuthenticatedIdentifierAction @Inject()(override val authConnector: AuthCo
...
     } else {
       authorised(AuthProviders(GovernmentGateway))
     }
   }
 
 }
-
-trait IdentifierAction extends ActionBuilder[IdentifierRequest] with ActionFunction[Request, IdentifierRequest]
-
-class SessionIdentifierAction @Inject()(config: FrontendAppConfig)
```
gokyo: not used

app/views/components/heading.scala.html
```diff
@@ -14,6 +14,6 @@
  * limitations under the License.
  *@
 
-@(headingKey: String, headingSize: String = "heading-xlarge")(implicit messages: Messages)
+@(key: String, headingSize: String = "heading-xlarge")(args: Any*)(implicit messages: Messages)
```
gokyo: I needed this trick because the heading for the EORI page needed an argument

app/views/components/heading.scala.html
```diff
@@ -14,6 +14,6 @@
  * limitations under the License.
  *@
 
-@(headingKey: String, headingSize: String = "heading-xlarge")(implicit messages: Messages)
+@(headingKey: String, headingSize: String = "heading-xlarge")(messageArgs: Any*)(implicit messages: Messages)
```
gokyo: I needed this trick because the heading for the EORI page needed an argument

app/views/confirmation.scala.html
```diff
@@ -24,7 +24,7 @@
     bodyClasses = None) {
 
     <div class="govuk-box-highlight">
-        @components.heading("confirmation.heading", "heading-xlarge")
```
gokyo: `heading-xlarge` is the default value in `heading.scala` - so need to specify it here

app/views/confirmation.scala.html
```diff
@@ -24,7 +24,7 @@
...
+        @components.heading("confirmation.heading")()
```
gokyo: `heading-xlarge` is the default value in `heading.scala` - so need to specify it here

app/views/components/heading.scala.html
ed0906: Instead of passing in `headerKey` and having to do hacks would it not make more sense to pass in the actual value we want as the header?

SteveSugden: I'd tend to agree with you Ed.

ed0906: e.g we would use it as `heading(messages(key))` rather than `heading(key)`

SteveSugden: Only arg for keeping it as is is the amount of refactoring of views and tests that would be involved.

ed0906: Looks like @gokyo already refactored them all to add the `()` to the end

SteveSugden: yes :-) .  He's just finished implementing your suggestion

PR #62, opened 2019-02-25
DIT-793: retrieving EORI details correctly

gokyo:


app/mapper/CaseRequestMapper.scala
```diff
@@ -59,17 +42,17 @@ class CaseRequestMapper {
...
 
     val app = Application(
-      holder = holder,
+      holder = holderDetails,
       contact = contact,
       agent = agentDetails,
       offline = false,
-      goodName = describeYourItem.map(_.field1).getOrElse("CaseRequestMapperTestN/A"),
+      goodName = describeYourItem.map(_.field1).getOrElse("N/A"),
       goodDescription = describeYourItem.map(_.field2).getOrElse("N/A"),
```
gokyo: Are we happy to save `N/A`?

app/mapper/CaseRequestMapper.scala
```diff
@@ -59,17 +42,17 @@ class CaseRequestMapper {
...
+    val agentDetails: Option[AgentDetails] = buildAgentDetails(userEori)(answers)
+    val holderDetails: EORIDetails = buildHolderDetails(userEori)(answers)
 
     val app = Application(
-      holder = holder,
+      holder = holderDetails,
       contact = contact,
       agent = agentDetails,
       offline = false,
-      goodName = describeYourItem.map(_.field1).getOrElse("CaseRequestMapperTestN/A"),
```
gokyo: Are we happy to save `N/A`?

app/mapper/CaseRequestMapper.scala
pllaurado: Maybe we could be more restrictive there by throwing and exception or stopping the process

pllaurado: Ultimately this is mandatory data that which is strictly necessary to finish the process

gokyo: Ok. I will throw an error

PR #74, opened 2019-03-02
Bau check your answers

SteveSugden:
Fix for problem of displaying agent details section when trader
Unit Test for CheckYourAnswersHelper to improve code coverage


app/viewmodels/RepeaterAnswerSection.scala
```diff
@@ -20,4 +20,6 @@ case class RepeaterAnswerSection (headingKey: String,
                                   relevanceRow: AnswerRow,
                                   rows: Seq[RepeaterAnswerRow],
                                   addLinkKey: String,
-                                  addLinkUrl: String) extends Section
+                                  addLinkUrl: String) extends Section {
+  override def isEmpty: Boolean = rows.isEmpty
```
gokyo: is the file name (and the class name) correct?
`RepeaterAnswerSection`?
Did you mean `Repeated`?

SteveSugden: I don't know.  It's an existing class.  Don't think we use it.

gokyo:
`rows` are used by both children of `Section`
You could define it on the parent trait and define the method `isEmtpy` (with a concrete implementation) there, so you avoid repeating the same code

app/viewmodels/RepeaterAnswerSection.scala
gokyo: You could fix this typo

gokyo: we could check what is the purpose of these classes and do some cleaning

test/unit/views/CheckYourAnswersViewSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+
+      text must include("Your details")
+      text must include("Trader Details")
+      text must not include("Details of the business, organisation or individual you represent")
+    }
+
+    "contain the answers for an agent" in {
+      val text = asDocument(createAgentView()).text()
+
+      text must include("Your details")
```
gokyo: is `include` the best way to assert? I think we should assert by equality (exact match)

app/views/check_your_answers.scala.html
```diff
@@ -31,7 +31,7 @@
 
         @components.heading(messages("checkYourAnswers.heading"))
 
-        @for(section <- answerSections){
+        @for(section <- answerSections if !section.isEmpty){
```
gokyo: you can make it more readable:
`if section.nonEmpty`

test/unit/views/CheckYourAnswersViewSpec.scala
SteveSugden: Exact match on a view?

gokyo: sorry, my fault. 
I did not realise you are comparing the whole HTML content

PR #75, opened 2019-03-05
Dit 802 application pdf

SteveSugden:


app/config/FrontendAppConfig.scala
```diff
@@ -48,6 +48,7 @@ class FrontendAppConfig @Inject() (override val runModeConfiguration: Configurat
   lazy val loginContinueUrl: String = loadConfig("urls.loginContinue")
   lazy val bindingTariffClassificationUrl: String = baseUrl("binding-tariff-classification")
   lazy val bindingTariffFileStoreUrl: String = baseUrl("binding-tariff-filestore")
+  lazy val pdfGeneratorUrl: String = baseUrl("pdf_generator_service")
```
ed0906: Minor but we usually go for `-` instead of `_` for these

ed0906: It will need to be added to `app-config-base` if you havent already

app/views/pdftemplates/applicationPdf.scala.html
```diff
@@ -0,0 +1,269 @@
...
+        <div class="centered-content">
+            <table class="header">
+                <tr>
+                    <td><img src="@createLink("images/hmrc-logo.jpg")" width="150" height="100"></td>
+                    <td> <h1 class="title">@messages("pdf.application.title")</h1> </td>
+                </tr>
+            </table>
+
+            <p>@messages("pdf.application.submitted"): @Dates.format(c.createdDate)</p>
+            <p>@messages("pdf.application.reference"): GB@c.reference</p>
```
ed0906: I actually think the GB but was put in the confirmation page by mistake, I think we should ask Matt if it’s meant to be there (and here)

I think it’s EORIs that start with GB

app/connectors/PdfGeneratorServiceConnector.scala
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package connectors
+
+import config.FrontendAppConfig
+import javax.inject.{Inject, Singleton}
+import play.api.libs.ws.{WSClient, WSResponse}
```
gokyo: We have to use `uk.gov.hmrc.play.bootstrap.http.HttpClient` rather than `WSClient`

gokyo: you need to use `http-verbs` - see the other connectors

app/controllers/PdfDownloadController.scala
```diff
@@ -0,0 +1,53 @@
...
+
+    caseService.getCaseForUser(request.eoriNumber, reference) flatMap {
+      case Some(c: Case) =>
+        fileService.getAttachmentMetadata(c) flatMap { f =>
+          pdfService.generatePdf(
+            s"confirmation_$reference.pdf",
+            applicationPdf(appConfig, c, f)
+          )
+        }
+      case _ => throw new Exception(s"Case ($reference) not found for user ${request.eoriNumber}")
```
gokyo: I think throwing an exception here (controller) is not the best way to handle errors.
You should return a 404 Not Found
See the other controllers we wrote

app/service/CasesService.scala
```diff
@@ -20,14 +20,27 @@ import connectors.BindingTariffClassificationConnector
...
-class CasesService @Inject()(connector: BindingTariffClassificationConnector){
+class CasesService @Inject()(connector: BindingTariffClassificationConnector) {
 
   def create(c: NewCaseRequest)(implicit hc: HeaderCarrier): Future[Case] = {
     connector.createCase(c)
   }
 
+  def getCaseForUser(userEori: String, reference: String)(implicit hc: HeaderCarrier): Future[Option[Case]] = {
+
+    connector.findCase(reference).map {
```
gokyo: I believe this code could be more concise

app/service/CasesService.scala
```diff
@@ -20,14 +20,27 @@ import connectors.BindingTariffClassificationConnector
...
+      case Some(c) if caseBelongsToUser(c, userEori) => Some(c)
+      case _ => None
+    }
+  }
+
+  private def caseBelongsToUser(c: Case, eoriNumber: String): Boolean = {
```
gokyo: this code is not very readable
Also the condition `c.application.holder.eori == eoriNumber` is written twice

app/service/PdfService.scala
```diff
@@ -0,0 +1,44 @@
...
+
+  def generatePdf(filename: String, htmlContent: Html): Future[Result] = {
+
+    connector.generatePdf(htmlContent).map {
+      response =>
+        response.status match {
+          case OK => Results.Ok(response.bodyAsBytes.toArray)
+            .as("application/pdf")
+            .withHeaders("Content-Disposition" -> s"attachment; filename=$filename")
+          case _ => Results.BadRequest(response.body)
```
gokyo: I think this code can be written better
You return 400 regardless of the upstream status (404, 500, etc)

app/controllers/PdfDownloadController.scala
ed0906: It’s probably the first use case for this pattern in the trader frontend. In the internal UI we just render a separate view that says “case no found”

app/service/CasesService.scala
ed0906: I think a ‘filter’ would work instead of a map

app/service/CasesService.scala
ed0906: Just a suggestion but we could add a method to ‘case’ called ‘hasEORI(eori: String):Boolean’

app/service/PdfService.scala
ed0906: Seems a little strange in my opinion to have a ‘Service’ that is handling responses and requests, shouldn’t that be left to the controller and connector?

app/config/FrontendAppConfig.scala
SteveSugden: Done

app/service/CasesService.scala
SteveSugden: I added your suggested method to the Case and have attempted to implement the filter - 
```connector.findCase(reference).filter(maybeCase => maybeCase.exists(_.hasEoriNumber(userEori)))```

This works until the case is filtered out (because it belongs to another Eori) and then I get `java.util.NoSuchElementException: Future.filter predicate is not satisfied`

File: app/service/CasesService.scala (hunk 6 above)
Comment (SteveSugden) [reply app/service/CasesService.scala]: Done

File: app/service/PdfService.scala (hunk 7 above)
Comment (SteveSugden) [reply app/service/PdfService.scala]: I agree.  I think it stems from using the WSClient in the connector.  Once I work out how to use the HttpClient I think I can get the connector to return something more useful - e.g. an Array[Byte] with which the service can create the Result.  Then (I'm guessing) I won't have to worry about bad responses - they'll just bubble up.

File: conf/application.conf [hunk 8]
```diff
@@ -104,6 +104,11 @@ microservice {
         port = 9583
       }
 
+      pdf_generator_service {
```
gokyo: for consistency with other services, use `-` rather than `_`

gokyo: have you added this dependency to our `service-manager-config` definitions?

gokyo: have you added this dependency to the runbook?

app/config/FrontendAppConfig.scala
gokyo: for consistency, I prefer we use `-` rather than `_`

app/connectors/PdfGeneratorServiceConnector.scala
```diff
@@ -0,0 +1,35 @@
...
+import play.twirl.api.Html
+
+import scala.concurrent.Future
+
+@Singleton
+class PdfGeneratorServiceConnector @Inject()(configuration: FrontendAppConfig, ws: WSClient) {
```
gokyo: remove `WSClient` - we need to use `uk.gov.hmrc.play.bootstrap.http.HttpClient`

app/connectors/PdfGeneratorServiceConnector.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  private lazy val url = s"${configuration.pdfGeneratorUrl}/pdf-generator-service/generate"
+
+  def generatePdf(html: Html): Future[WSResponse] = {
+    ws.url(url).post(Map("html" -> Seq(html.toString)))
```
gokyo: use `uk.gov.hmrc.play.bootstrap.http.HttpClient`

app/controllers/PdfDownloadController.scala
gokyo: Yes, I meant let's check also other microservices controllers

test/unit/connectors/PdfGeneratorServiceConnectorSpec.scala
```diff
@@ -0,0 +1,69 @@
...
+import uk.gov.hmrc.play.bootstrap.http.DefaultHttpClient
+import uk.gov.hmrc.play.test.{UnitSpec, WithFakeApplication}
+
+class PdfGeneratorServiceConnectorSpec extends UnitSpec with WithFakeApplication
+  with WiremockTestServer with MockitoSugar with BeforeAndAfterEach with ResourceFiles {
+
+  private val config = mock[FrontendAppConfig]
+  private val pdfTemplate = mock[Html]
+  private val actorSystem = ActorSystem.create("test")
+  private val wsClient: WSClient = fakeApplication.injector.instanceOf[WSClient]
```
gokyo: remove

app/views/pdftemplates/applicationPdf.scala.html
gokyo: yes, we should remove `GB`

app/service/PdfService.scala
gokyo: yes, do not handle errors here, do it in the controller

check also the operational UI in case

app/service/CasesService.scala
gokyo: did you commit and push?
I cannot see the change

gokyo: Can you extract `c.application.holder.eori == eoriNumber` to a boolean method?

SteveSugden: Yes - forgot to push last night

conf/application.conf
SteveSugden: No.  I wanted to see if it would run "out of the box" with no config changes.   I had an experiment with sm config but there appears to be no way to change config for a service.

SteveSugden: What's that (runbook)?

app/views/pdftemplates/applicationPdf.scala.html
SteveSugden: 'GB' removed from pdf and confirmation page.  I retained the use of messages to format the reference so it can be easily changed back if required.

app/service/CasesService.scala
ed0906: I was thinking it would just become 
`connector.findCase(reference).filter(_.filter(_.hasEori(userEori)))`

ed0906: I take it back, I think you need to keep the `map` but you can use a `filter` inside it

app/service/PdfService.scala
ed0906: > I agree. I think it stems from using the WSClient in the connector. Once I work out how to use the HttpClient I think I can get the connector to return something more useful - e.g. an Array[Byte] with which the service can create the Result. Then (I'm guessing) I won't have to worry about bad responses - they'll just bubble up.

I think using the http verbs client like the other connectors use should make it easier as you can specify a return type e.g. `client.GET[SomePDFType](url)`

app/service/CasesService.scala
SteveSugden: That form works (the first one).  Thanks

app/connectors/PdfGeneratorServiceConnector.scala
SteveSugden: Tried and failed.  We threw in the towel.

conf/application.conf
SteveSugden: PR created and merged

SteveSugden: Done

app/service/PdfService.scala
```diff
@@ -0,0 +1,41 @@
...
+import scala.concurrent.Future
+
+@Singleton
+class PdfService @Inject()(connector: PdfGeneratorServiceConnector) {
+
+  def generatePdf(htmlContent: Html): Future[BinaryFile] = {
+
+    connector.generatePdf(htmlContent).map {
+      response =>
+        response.status match {
```
ed0906: I still feel like this logic belongs in the connector

app/connectors/PdfGeneratorServiceConnector.scala
ed0906: @SteveSugden I'm happy to help with using http verbs if you like.

The reason It's important (I believe) is because it does a bunch of auditing and metrics for each request under the hood + passes through any mdtp auth tokens and request IDs etc

conf/application.conf
SteveSugden: Done - https://confluence.tools.tax.service.gov.uk/display/BTI/BTI+Application+Service+Runbook

app/controllers/PdfDownloadController.scala
```diff
@@ -0,0 +1,53 @@
...
+    caseService.getCaseForUser(request.eoriNumber, reference) flatMap {
+      case Some(c: Case) =>
+        fileService.getAttachmentMetadata(c) flatMap { attachmentData =>
+          pdfService.generatePdf(applicationPdf(appConfig, c, attachmentData)).map({ binaryFile =>
+            Results.Ok(binaryFile.content)
+              .as(binaryFile.contentType)
+              .withHeaders("Content-Disposition" -> s"attachment; filename=confirmation_$reference.pdf")
+          })
+        }
+      case _ => throw new Exception(s"Case ($reference) not found for user ${request.eoriNumber}")
```
gokyo: I think we can use the same approach of the operational UI (showing the "case not found error page") or alternatively this is ok, but we should not log the request to kibana because it contains PII (eori number)

app/connectors/PdfGeneratorServiceConnector.scala
```diff
@@ -0,0 +1,44 @@
...
+class PdfGeneratorServiceConnector @Inject()(configuration: FrontendAppConfig, ws: WSClient) {
+
+  private lazy val url = s"${configuration.pdfGeneratorUrl}/pdf-generator-service/generate"
+
+  def generatePdf(html: Html): Future[BinaryFile] = {
+    ws.url(url).post(Map("html" -> Seq(html.toString))) map {
+      response =>
+        response.status match {
+          case OK => BinaryFile(contentType = "application/pdf", content = response.bodyAsBytes.toArray)
+          case _ => throw new Exception(s"Error calling PdfGeneratorService $response")
```
gokyo: do not throw exception like this
return instead a `Future.failed`

gokyo:
```
ws.url(url).post(Map("html" -> Seq(html.toString))) flatMap {
      response =>
        response.status match {
          case OK => Future.successful(BinaryFile(contentType = "application/pdf", content = response.bodyAsBytes.toArray))
          case _ => Future.failed(new Exception(s"Error calling PdfGeneratorService $response"))
```

app/connectors/PdfGeneratorServiceConnector.scala
```diff
@@ -0,0 +1,44 @@
...
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+@Singleton
+class PdfGeneratorServiceConnector @Inject()(configuration: FrontendAppConfig, ws: WSClient) {
+
+  private lazy val url = s"${configuration.pdfGeneratorUrl}/pdf-generator-service/generate"
+
+  def generatePdf(html: Html): Future[BinaryFile] = {
```
pllaurado: I would return a Future[Option[PdfFile]] an avoid to throw this Exception.

app/connectors/PdfGeneratorServiceConnector.scala
```diff
@@ -0,0 +1,44 @@
...
+    ws.url(url).post(Map("html" -> Seq(html.toString))) map {
+      response =>
+        response.status match {
+          case OK => BinaryFile(contentType = "application/pdf", content = response.bodyAsBytes.toArray)
```
pllaurado: BinaryFile looks to generic to me. Maybe PdfFile or something that describe quickly what are we doing

app/connectors/PdfGeneratorServiceConnector.scala
pllaurado: I would avoid this one

app/connectors/PdfGeneratorServiceConnector.scala
SteveSugden: I'm going to return a Future.failed (after discussion with @gokyo ) - an optional pdf is not much use to the caller

app/connectors/PdfGeneratorServiceConnector.scala
SteveSugden: Avoid what ?

ed0906: Looks fine to me, we need to handle it somewhere and this seems like the right place

app/connectors/PdfGeneratorServiceConnector.scala
ed0906: I think given it contains a content type it's fine

app/connectors/PdfGeneratorServiceConnector.scala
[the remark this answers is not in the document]
SteveSugden: We've had 2 of our finest trying to get this to work.   We gave up and are doing the same as every other user of the pdf-generator-service

app/service/CasesService.scala
```diff
@@ -20,14 +20,27 @@ import connectors.BindingTariffClassificationConnector
...
+    c.application.agent.map(agent => agent.eoriDetails.eori == eoriNumber || c.application.holder.eori == eoriNumber)
```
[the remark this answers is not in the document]
SteveSugden: You are looking at outdated code.  That method is on the Case now

app/service/PdfService.scala
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package service
+
+import connectors.PdfGeneratorServiceConnector
+import javax.inject.{Inject, Singleton}
+import play.api.http.Status._
+import play.api.mvc.{Result, Results}
```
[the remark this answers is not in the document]
SteveSugden: Again - you are looking at a previous iteration

PR #80, opened 2019-03-07
Dit 81 ruling pdf

SteveSugden:
Adds endpoint to generate Ruling PDF for a case that has a decision.  
Updates the links on the Ruling tab on the trader home page.
Some code tidy (fix warnings) in recently merged files


app/controllers/PdfDownloadController.scala
```diff
@@ -42,10 +42,20 @@ class PdfDownloadController @Inject()(appConfig: FrontendAppConfig,
         pdfService.generatePdf(applicationPdf(appConfig, c, attachmentData)).map { binaryFile =>
           Results.Ok(binaryFile.content)
             .as(binaryFile.contentType)
-            .withHeaders(CONTENT_DISPOSITION -> s"filename=confirmation_$reference.pdf")
+            .withHeaders(CONTENT_DISPOSITION -> s"filename=BTIConfirmation$reference.pdf")
         }
       }
     }
   }
```
gokyo: The two methods in this class are very similar.
You could create a common private method used by both (DRY principle)

app/models/Case.scala
```diff
@@ -35,11 +35,12 @@ case class Case
...
-  def hasEoriNumber(eoriNumber: String): Boolean = {
-    application.holder.eori == eoriNumber || application.agent.exists(_.eoriDetails.eori == eoriNumber)
-  }
+  def hasEoriNumber(eoriNumber: String): Boolean = application.holder.eori == eoriNumber || application.agent.exists(_.eoriDetails.eori == eoriNumber)
 
   def hasActiveDecision: Boolean = this.decision.flatMap(_.effectiveEndDate).exists(_.compareTo(Instant.now) >= 0)
 
   def hasExpiredDecision: Boolean = this.decision.flatMap(_.effectiveEndDate).exists(_.compareTo(Instant.now) < 0)
+
+  def hasRuling: Boolean = decision.exists(d => d.effectiveStartDate.isDefined && d.effectiveEndDate.isDefined)
```
gokyo: is this the proper way to make sure there is a ruling ?

app/controllers/PdfDownloadController.scala
SteveSugden: Pulled out the common code into a private method

app/models/Case.scala
SteveSugden: Not sure if that is the "complete" definition of "case has a ruling".    It's good enough (IMO) to ensure that a ruling cert pdf will have all of the information it requires (i.e. a Decision block with start and end dates).

The only other additional check I can think of is that the case status is Completed or Canceled.  I'll check with Matt/Joe

pllaurado: Those are my acceptance criteria in DIT-79 about what to show as a Ruling : 

> Ruling status mappings must be presented as follows:
> "Active" where Case.Status = Complete AND Current Date <= ruling End Date
> "Expired" where Case.Status = Complete and Current Date > ruling End Date
> "Cancelled" where Case.Status = Cancelled

SteveSugden: Thanks - I've added a question on my ticket - https://jira.tools.tax.service.gov.uk/browse/DIT-81

app/controllers/PdfDownloadController.scala
```diff
@@ -42,10 +42,20 @@ class PdfDownloadController @Inject()(appConfig: FrontendAppConfig,
         pdfService.generatePdf(applicationPdf(appConfig, c, attachmentData)).map { binaryFile =>
           Results.Ok(binaryFile.content)
             .as(binaryFile.contentType)
-            .withHeaders(CONTENT_DISPOSITION -> s"filename=confirmation_$reference.pdf")
+            .withHeaders(CONTENT_DISPOSITION -> s"filename=BTIConfirmation$reference.pdf")
```
pllaurado: I don't which are the standards / conventions  for file name. Is this name ok ? BTIConfirmation$reference.pdf

SteveSugden: I checked on the file names with Matt yesterday.   He preferred / was happy with the new format

app/models/Case.scala
gokyo: @SteveSugden 
as agreed with @pllaurado , we would like this method to assert the case status (you can keep the dates assertion, but perhaps better to get rid of it and just check the case status == completed / decision_made)

SteveSugden: Ok.  But it will have to be status = Completed or Cancelled

PR #85, opened 2019-03-14
DIT-4881 Make EORI editable depending on CDS Flag

ed0906:
I've implemented it as follows:
- The existing CDS bypass flag in config decides wether its permitted for someone to not have an EORI

After this point the flag isn't used and everything hangs off whether the EORI is present on the request, which has involved:

- Making the Request EORI optional
- Making the main journey toggle content based on whether the request EORI is present


ed0906:
Dont merge until It's been tested etc :)

ed0906:
@jumpertester The IDs on the Register Business For EORI page may have changed (from `field1`, `field2`) to something more meaningful e.g. `businessName`, `addressLine1`. I imagine this will cause a few test failures

conf/application.conf
```diff
@@ -169,7 +169,7 @@ filters {
   }
 }
 
-cdsEnrolmentCheckEnabled = true
+cdsEnrolmentCheckEnabled = false
```
ed0906: This is overridden in production

app/models/WhichBestDescribesYou.scala
```diff
@@ -48,7 +48,7 @@ object WhichBestDescribesYou {
     }
   }
 
-  def isBusinessRepresentative(answers: UserAnswers): Boolean = {
+  def theUserIsAnAgent(answers: UserAnswers): Boolean = {
```
gokyo: `isSignedInUserAnAgent`

app/mapper/CaseRequestMapper.scala
```diff
@@ -105,15 +99,15 @@ class CaseRequestMapper {
...
 
-  private def buildEoriDetailsFromCurrentUser(eoriNumber: String): RegisteredAddressForEori => EORIDetails = { details: RegisteredAddressForEori =>
+  private def toEoriDetails(details: RegisteredAddressForEori): EORIDetails = {
     EORIDetails(
-      eoriNumber,
-      details.field1, // business name
-      details.field2, // address line 1
-      details.field3, // address line 2 (town)
+      details.eori,
+      details.businessName, // business name
```
gokyo: The comments were useful before, when using field1, field2, etc
Now that you named the fields with the correct names, the comments can go away

app/models/WhichBestDescribesYou.scala
ed0906: `if(is...)` never reads very well in my opinion

app/mapper/CaseRequestMapper.scala
ed0906: Removed

ed0906:
Merged as part of #87

PR #86, opened 2019-03-14
DIT-860: fixing bug

gokyo:


test/unit/views/SupportingMaterialFileListViewSpec.scala
```diff
@@ -17,26 +17,73 @@
...
+    supportingMaterialFileList(frontendAppConfig, f, files, NormalMode)(fakeRequest, messages)
+  }
 
   "SupportingMaterialFileList view" must {
 
     behave like normalPage(createView, messageKeyPrefix)()
 
     behave like pageWithBackLink(createView)
 
+//    behave like yesNoPage() // TODO: not working at the moment as it has not been created as the other Yes/No pages
```
gokyo: @pllaurado @SteveSugden 
I feel like the refactoring has not done properly for this page.
I am waiting for your opinion here.

SteveSugden: I think Pol's opinion is more valid than mine.   I suspect the reason is that this is not a simple "Yes/No goto next page" page.

gokyo: It is a Yes/No behaviour.
The endpoint is: `/supportingMaterialFileList`

it is not a simple page, but we did not defined it properly as the other boolean questions...

Also look here: https://github.com/hmrc/binding-tariff-trader-frontend/blob/master/test/unit/views/SupportingInformationViewSpec.scala#L26

It says: `SupportingInformationViewSpec extends YesNoViewBehaviours`

PR #90, opened 2019-03-21
BAU: cleaning code

gokyo:


app/controllers/PdfDownloadController.scala
```diff
@@ -39,47 +40,49 @@ class PdfDownloadController @Inject()(appConfig: FrontendAppConfig,
...
-        case Some(eori) => getApplicationPDF(eori, reference)
-        case None => Future.successful(Redirect(controllers.routes.SessionExpiredController.onPageLoad()))
-      }
-    } else {
-      Future.successful(Redirect(controllers.routes.UnauthorisedController.onPageLoad()))
-    }
+    getPdf(reference, token, getApplicationPDF)
   }
 
   def ruling(reference: String, token: Option[String]): Action[AnyContent] = identify.async { implicit request =>
```
gokyo: there were two methods doing exactly the same thing... so I created a generic method used by both

app/controllers/PdfDownloadController.scala
```diff
@@ -39,47 +40,49 @@ class PdfDownloadController @Inject()(appConfig: FrontendAppConfig,
...
+  private def getRulingPDF(implicit request: Request[AnyContent]): (String, String) => Future[Result] = { (eori, reference) =>
+    caseService.getCaseWithRulingForUser(eori, reference) flatMap { c: Case =>
+      generatePdf(
+        rulingPdf(appConfig, c, c.decision.getOrElse(throw new IllegalStateException("Missing decision"))), s"BTIRuling$reference.pdf"
+      )
     }
   }
 
-  private def getApplicationPDF(eori: String, reference: String)(implicit r: Request[_]): Future[Result] = {
+  private def getApplicationPDF(implicit request: Request[AnyContent]): (String, String) => Future[Result] = { (eori, reference) =>
```
ed0906: I think the signature of this was better before. Its not very intuitive now

app/controllers/PdfDownloadController.scala
```diff
@@ -39,47 +40,49 @@ class PdfDownloadController @Inject()(appConfig: FrontendAppConfig,
...
+        case Some(eori) => pdf(eori, reference)
         case None => Future.successful(Redirect(controllers.routes.SessionExpiredController.onPageLoad()))
       }
-    } else {
-      Future.successful(Redirect(controllers.routes.UnauthorisedController.onPageLoad()))
+      case _ => Future.successful(Redirect(controllers.routes.UnauthorisedController.onPageLoad()))
+    }
+  }
+
+  private def getRulingPDF(implicit request: Request[AnyContent]): (String, String) => Future[Result] = { (eori, reference) =>
```
ed0906: I think the signature of this was better before. Its not very intuitive now

app/controllers/PdfDownloadController.scala
gokyo: I know, but I need a function if I want to remove repetitions from the code

gokyo: what do u suggest?

ed0906: It is a function, you can just change the way it is called.

gokyo: How do you like it?

gokyo:
@ed0906 
Do you like it more now?

PR #96, opened 2019-03-27
Dit 887

SteveSugden:


app/views/describeYourItem.scala.html
```diff
@@ -36,21 +36,32 @@
...
-            field = form("field1"),
-            label = messages("describeYourItem.field1")
+            field = form("name"),
+            label = messages("describeYourItem.name")
         )
 
         @components.input_textarea(
-            field = form("field2"),
-            label = messages("describeYourItem.field2"),
+            field = form("description"),
```
ed0906: Nice one for doing this, just be wary the ids are generated using the field name so a few ATs may be broken

SteveSugden: We will mop up during the fixing of the TraderUI ATs.   (Or Ash will)

PR #101, opened 2019-04-10
DIT-967 import export new question

pllaurado:


app/controllers/ImportOrExportController.scala
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    Ok(importOrExport(appConfig, preparedForm, mode))
+  }
+
+  def onSubmit(mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData).async { implicit request =>
+
+    form.bindFromRequest().fold(
+      (formWithErrors: Form[_]) =>
+        Future.successful(BadRequest(importOrExport(appConfig, formWithErrors, mode))),
```
ed0906: BadRequest? we don't usually do that for form errors do we?

pllaurado: Common pattern in all the app. Scaffold behaviours

PR #102, opened 2019-04-29
Dit 976

aaroncarroll82:
Draft until unit tests added


app/views/components/table_rulings.scala.html
```diff
@@ -58,7 +58,7 @@
                 <tbody id="rulings-list-body">
                     @for((c: Case, index: Int) <- cases.results.zipWithIndex) {
                         <tr id="rulings-list-row-@{index}">
-                            <td id="rulings-list-row-@{index}-reference">@c.reference</td>
+                            <td id="rulings-list-row-@{index}-reference">GB@c.reference</td>
```
ed0906: We need to double check with Joe if GB should be part of the reference. We have removed it from lots of places in the past but I know Steve P still has it in his prototypes

app/views/pdftemplates/rulingPdf.scala.html
```diff
@@ -103,7 +103,7 @@ <h2>@messages("pdf.ruling.section.ruling.heading")</h2>
                 </p>
                 <p>
                     <span class="section-label">@messages("pdf.ruling.section.ruling.rulingReference.label"):</span>
-                    @messages("caseReference.format", c.reference)
+                    GB@messages("caseReference.format", c.reference)
```
ed0906: Similar here. Has Joe said it should start with GB?

app/views/ruling_information.scala.html
```diff
@@ -0,0 +1,104 @@
...
+@(appConfig: FrontendAppConfig, c: Case)(implicit request: Request[_], messages: Messages)
+
+    @main_template(
+        title = s"${messages("rulingInformation.title")}",
+        appConfig = appConfig,
+        bodyClasses = None)  {
+
+        @components.back_link()
+
+        @components.heading(s"${messages("rulingInformation.Heading")} GB${c.reference}")
```
ed0906: Similar here

app/views/ruling_information.scala.html
```diff
@@ -0,0 +1,104 @@
...
+            <div class="grid-full-width">
+                <div class="column-one-third">@messages("rulingInformation.Holder")</div>
+                <div class="column-two-thirds">@c.application.holder.businessName</div>
+            </div>
+        </div>
+        </br>
+        <div class="grid-row">
+            <div class="grid-full-width">
+                <div class="column-one-third">@messages("rulingInformation.Ruling")</div>
+                <div class="column-two-thirds">GB@c.reference</div>
```
ed0906: Similar here

app/views/components/table_rulings.scala.html
SteveSugden: I thought we stop prefixing the reference with "GB" a while back.    I presume that re-introduction is deliberate?

app/views/ruling_information.scala.html
```diff
@@ -0,0 +1,104 @@
...
+            </div>
+        </div>
+        </br>
+        <div class="grid-row">
+            <div class="grid-full-width">
+                <div class="column-one-third">@messages("rulingInformation.StartDate")</div>
```
ed0906: We generally use camel case (`StartDate`)

app/views/ruling_information.scala.html
```diff
@@ -0,0 +1,104 @@
...
+                <div class="column-two-thirds">@Dates.format(c.decision.get.effectiveStartDate)</div>
+            </div>
+        </div>
+        </br>
+        <div class="grid-row">
+            <div class="grid-full-width">
+                <div class="column-one-third">@messages("rulingInformation.ExpiryDate")</div>
```
ed0906: We generally use camel case (`EndDate`)

conf/messages.en
```diff
@@ -465,4 +465,25 @@ case.ruling.status.cancelled = Cancelled
 case.ruling.status.active = Active
 case.ruling.status.expired = Expired
 
-case.ruling.downloadPdf = Download pdf
+case.ruling.downloadPdf = View BTI Ruling
+
+rulingInformation.title = Information on Ruling
+rulingInformation.Heading = Important information about BTI ruling
```
ed0906: We appear to have changed the casing style, lets be consistent with what was already there

test/unit/views/RulingInformationViewSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+    "show the expected element values" in {
+
+      val doc = asDocument(createView(rulingCase).apply())
+      assertContainsText(doc,"GB"+rulingCase.reference)
+      assertContainsText(doc,rulingCase.application.holder.businessName)
+      assertContainsText(doc,rulingCase.application.holder.businessName)
+      assertContainsText(doc,Dates.format(ruling.effectiveStartDate))
+      assertContainsText(doc,Dates.format(ruling.effectiveEndDate))
+
+      assertContainsText(doc,messages("rulingInformation.CommodityIntro")+" "+ ruling.bindingCommodityCode+".")
```
SteveSugden: You could have a go at string formatting the Scala way using s"....."

aaroncarroll82: Old habits die hard... Updated to scala string formatting now

conf/messages.en
aaroncarroll82: There's a few different styles used throughout the messages file, but you are right - here there's an inconsistency that means that of the latest entry these don't even all match each other. I'll update to be proper camelcase

app/controllers/RulingController.scala
```diff
@@ -0,0 +1,49 @@
...
+
+  def viewRuling(reference: String): Action[AnyContent] = identify.async { implicit request =>
+
+    request.eoriNumber match {
+      case Some(eori: String) =>
+        service.getCaseForUser(eori,reference) flatMap {
+          c: Case => successful(Ok(ruling_information(appConfig, c)))
+        }
+
+      case None => successful(Redirect(routes.BeforeYouStartController.onPageLoad()))
```
pllaurado: Which is the actual behaviour for this??   Looks like the user can request a BTI but suddenly appears to be in the first page. I would rather have something more self-explanatory but I guess is more a design question

SteveSugden:
We used to do the "GBref" stuff as a parameterised value in the messages file

app/controllers/RulingController.scala
ed0906: This would be if you changed the URL to a reference that doesn't exist

app/views/components/table_rulings.scala.html
```diff
@@ -72,7 +72,7 @@
                                 }
                             </td>
                             <td id="rulings-list-row-@{index}-download">
-                                <a id="rulings-list-row-@{index}-download-link" target="_blank" href="@routes.PdfDownloadController.ruling(c.reference)">@messages("case.ruling.downloadPdf")</a>
+                                <a id="rulings-list-row-@{index}-download-link" target="_self" href="@routes.RulingController.viewRuling(c.reference)">@messages("case.ruling.downloadPdf")</a>
```
ed0906: Is target `_self` not the same as not having the target attribute? It can probably be removed

aaroncarroll82: I think it's default dependant on which browser is used, but modern browsers all use it as default so I can safely remove here. I'll get rid as part of tidy up.

app/views/ruling_information.scala.html
aaroncarroll82: Updated as per the other comment on messages references

app/views/ruling_information.scala.html
aaroncarroll82: Updated as per the other comment on messages references

app/views/components/table_rulings.scala.html
aaroncarroll82: After post-amigos this has been removed now, so no longer a thing

app/views/pdftemplates/rulingPdf.scala.html
aaroncarroll82: After post-amigos this has been removed now, so no longer a thing

app/views/ruling_information.scala.html
aaroncarroll82: After post-amigos this has been removed now, so no longer a thing

app/views/ruling_information.scala.html
aaroncarroll82: After post-amigos this has been removed now, so no longer a thing

app/views/components/table_rulings.scala.html
aaroncarroll82: After post-amigos this has been removed now, so no longer a thing

aaroncarroll82:
Removed the GB ref stuff after post amigos, so the general comments on this PR should all have resolved themselves in light of this.

PR #111, opened 2019-06-19
DIT-1054 - textarea DAC label 

pllaurado:


app/views/components/input_textarea.scala.html
```diff
@@ -26,27 +26,31 @@
...
-        @field.errors.map { error =>
-            <span class="error-message" id="error-message-@{field.id}-input">@messages(error.message, error.args: _*)</span>
-        }
-    </label>
+
+    @label.map { label =>
+        <label class="form-label @labelClass" for="@{field.id}">@label</label>
+    }
+    @labelHtml.map { label =>
+        <label for="@{field.id}">@label</label>
```
SteveSugden: Is this not going to result in two <labels>?  Before labelHtml was the content for a label; now it's a whole other <label>.  Or am I missing something?

PR #120, opened 2019-11-06
DIT-1186 | content changes for agent trader view

gorkemunsal:


app/views/templates/applicationTemplate.scala.html
```diff
@@ -28,17 +28,17 @@
...
 }
 
 <html>
-    <head>
-        <style>
+<head>
+    <style>
                 body {
                     font-size: 21px;
                     font-family: "nta", Arial, sans-serif;
```
vajarova: Is there a reason we have the styling within the view?

JoeSimmonds: This is the page used to drive the PDF generation so all of the assets need to be in the source I think

PR #162, opened 2020-06-18
DIT-1845 added new class for the postcode field to keep width as 33%

shiromi:


DavidGregory084:
@shiromi you shouldn't need to do a commit to re-run the build BTW, you can just close and reopen the PR

frikit:
life hack for me as well :D didn't think about this one 👍

shiromi:
> @shiromi you shouldn't need to do a commit to re-run the build BTW, you can just close and reopen the PR

@DavidGregory084 build failed , that's why committed again :(

PR #174, opened 2020-07-13
DIT-2154 Changes from Application Service content review

DavidGregory084:
Notes:
* The notice 600 page messages have not been changed yet because the final structure and content is still under discussion
* Same situation with the "Unauthorised" page which still refers to CDS
* The PDF content will be updated in another ticket


frikit:
Looks good

DavidGregory084:
Marking as not for merge for now because I think it will require changes to the journey tests

PR #180, opened 2020-08-04
DIT-1828 updated new contents from spreadsheet on Accessibility statement

shiromi:


shiromi:
there is merge conflicts to be resolve. will create a new PR

PR #183, opened 2020-08-12
DIT-1908 Add 'Error' prefix for screen readers for Application service

shiromi:


app/views/components/error_summary.scala.html
```diff
@@ -24,7 +24,7 @@ <h2 class="heading-medium error-summary-heading" id="error-summary-heading">
 
         <ul role="list" class="error-summary-list">
             @for(error <- errors) {
-                <li><a href="#@{error.key}">@messages(error.message, error.args:_*)</a></li>
+                <li><a href="#@{error.key}"><span class="visually-hidden">@messages("error.browser.title.prefix")</span>@messages(error.message, error.args:_*)</a></li>
```
adamliptrot-oc, line 27: The prefix shouldn't be added to messages inside the error summary, only against error messages next to fields.

shiromi: sure. thanks and I will fix it.

PR #190, opened 2020-08-25
DIT-2270 | Create case event added

gorkemunsal:


test/unit/controllers/DeclarationControllerSpec.scala
```diff
@@ -119,15 +123,46 @@ class DeclarationControllerSpec extends ControllerSpecBase with BeforeAndAfterEa
...
+      givenTheCaseCreatesSuccessfully()
+      givenTheCaseCreatedEventIsSuccessful()
+
+      val c = controller(extractDataFromCache)
+      val req = fakeRequest
+
+      await(c.onSubmit(NormalMode)(req))
+
+      verify(casesService, times(1)).addCaseCreatedEvent(
+        any[Case], any[Operator])(any[HeaderCarrier])
```
OC-Pomaroff, line 144: I don't think we have actually tested that we are creating the event with correct fields.

If we define a contact name,
```
when(contact.name).thenReturn("luigi")
```

then instead of `any[Case]` and `any[Operator]` we can ensure it is using the correct fields.

```suggestion
        refEq(createdCase), refEq(Operator("", Some("luigi"))))(any[HeaderCarrier])
```

app/service/CasesService.scala
```diff
@@ -70,4 +70,12 @@ class CasesService @Inject()(connector: BindingTariffClassificationConnector, em
     }
   }
 
+  def addCaseCreatedEvent(atar: Case, operator: Operator)
+                                 (implicit hc: HeaderCarrier): Future[Unit] = {
+    val event = NewEventRequest(CaseCreated("Application submitted"), operator)
```
OC-Pomaroff, line 75: Similar to what David suggested here https://github.com/hmrc/tariff-classification-frontend/pull/379#discussion_r477148443

It might be good to add/use the `addEvent` private function here as well (even though this will be the only consumer in this project for now, it makes it a bit clearer).

OC-Pomaroff (approved):
Looks good

kuldeepwadhwa9 (approved):

PR #196, opened 2020-09-07
DIT-2287 Add metrics collection code

DavidGregory084:
This PR adds a `HasMetrics` trait for use in services and connectors inspired by [similar code](https://github.com/hmrc/pertax-frontend/blob/master/app/metrics/HasMetrics.scala) in pertax-frontend.

I have changed the code somewhat so that it is harder to leak timers by forgetting to stop them or by failing to handle exceptions in the code block that is being executed.

The approach I have used is to try and ensure that the `completeWithSuccess` and `completeWithFailure` methods on the timer are idempotent, that is, whatever code stops the timer first should win and further attempts to stop it should do nothing.

The provided combinators then stop the timer on completion of the Future block regardless of whether the user did so.

### DIFF
--- a/app/metrics/HasMetrics.scala
+++ b/app/metrics/HasMetrics.scala
@@ -61,14 +61,14 @@ trait HasMetrics {
 
     def completeWithSuccess(): Unit = {
       if (timerRunning.compareAndSet(true, false)) {
-        timer.stop()
+        localMetrics.stopTimer(timer)
         localMetrics.incrementSuccessCounter(metric)
       }
     }
 
     def completeWithFailure(): Unit = {
       if (timerRunning.compareAndSet(true, false)) {
-        timer.stop()
+        localMetrics.stopTimer(timer)
         localMetrics.incrementFailedCounter(metric)
       }
     }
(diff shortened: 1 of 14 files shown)


app/metrics/HasMetrics.scala
```diff
@@ -0,0 +1,130 @@
...
+          timer.completeWithSuccess()
+      }
+
+      // Clean up timer for unhandled exceptions
+      result.failed.foreach { _ =>
+        timer.completeWithFailure()
+      }
+
+      result
+    }
```
DavidGregory084, line 87: Out of an abundance of caution, it might be a good idea to add a `try..finally` block around the call to `block` in case the user causes an exception while constructing the Future

app/metrics/HasMetrics.scala
```diff
@@ -0,0 +1,130 @@
...
+        timer.completeWithSuccess()
+      }
+
+      // Clean up timer on unhandled exceptions
+      result.failed.foreach { _ =>
+        timer.completeWithFailure()
+      }
+
+      result
+    }
```
DavidGregory084, line 122: Same here

gorkemunsal (approved):

PR #199, opened 2020-09-10
DIT-2267 | Added AddConfidentialInformation page

periklas:


kuldeepwadhwa9:
@periklas @gorkemunsal  Can you update the elements for Goods Name and Goods Description fields -
from "value" to "goodsName" and "goodsDescription" please?

PR #201, opened 2020-09-15
DIT-2268 | provide confidential information page

gorkemunsal:


app/controllers/ProvideConfidentialInformationController.scala
```diff
@@ -0,0 +1,76 @@
...
+                                        formProvider: ProvideConfidentialInformationFormProvider,
+                                        val provide_confidential_information: views.html.provideConfidentialInformation,
+                                        cc: MessagesControllerComponents
+                                     )(implicit ec:ExecutionContext) extends FrontendController(cc) with I18nSupport {
+
+  val form = formProvider()
+
+  def onPageLoad(mode: Mode) = (identify andThen getData andThen requireData) {
+    implicit request =>
+      val goodsName = request.userAnswers.get(ProvideGoodsNamePage).getOrElse("")
```
OC-Pomaroff, line 51: I wonder what is the best default here - should we default it to `.getOrElse("goods")` as per the AddConfidentialInformationController?

In ProvideGoodsDescriptionController there is no default and it would throw an exception. It might be worth checking all of these places and using the same default, maybe we could even move this logic into a Helper class or something?

OC-Pomaroff (approved):
Looks good

app/controllers/ProvideConfidentialInformationController.scala
DavidGregory084: I think the `getOrElse("goods")` is probably the best approach as the page text still makes sense that way 👍

gorkemunsal: this has now been addressed

PR #204, opened 2020-09-16
DIT-2309 # content changes to are you sending samples.

tipperba:


DavidGregory084:
This looks good to me @tipperba but I worked on it with you so I've added @colesadam to the review 😄

conf/messages.en
```diff
@@ -292,19 +292,18 @@ commodityCodeDigits.error.required = Enter a commodity code
 commodityCodeDigits.error.length = Commodity code must be 100 characters or less
 
 whenToSendSample.title = Sending samples to HMRC - Manage your Advance Tariff Rulings - GOV.UK
```
DavidGregory084, line 294: Let's check with Eszter whether we need to change the page title as well

tipperba: Have amended the title in line with slack conversation: 

D:
Hey chaps, not sure if Matt has mentioned it but he has spoken to Eszter about the page titles thing we looked at yesterday and apparently the page heading must contain the goods name but the page title should not contain it (rather just saying "goods").

B: 
Presumably it’s just the first part of the title that will change - we will retain the rest as is? So this:
Sending samples to HMRC - Manage your Advance Tariff Rulings - GOV.UK
becomes this:
Will you send a sample of the goods to HMRC? - Manage your Advance Tariff Rulings - GOV.UK

colesadam (approved):

PR #207, opened 2020-09-17
DIT-2312 Update content on the have you found a commodity code page

warrenskiwork1979:


OC-Pomaroff (approved):
LGTM

app/views/commodityCodeBestMatch.scala.html
```diff
@@ -35,10 +35,12 @@
...
             field = form("value"),
-            labelHtml = Some(components.heading(messages("commodityCodeBestMatch.heading"))),
+            labelHtml = Some(components.heading(
+                messages("commodityCodeBestMatch.heading", goodsName),
+                captionMsg = Some(messages("commodityCodeBestMatch.caption")))),
             hint = Some(messages("commodityCodeBestMatch.bodyText"))
         )
 
-        @components.submit_button()
+        @components.submit_button(Some(messages("commodityCodeBestMatch.submitLabel")))
```
OC-Pomaroff, line 44: Just noticed that we should probably use "site.save_and_continue"

DavidGregory084: Yeah we were talking about this a bit yesterday - we thought we could go through the site when we did the navigation part and change all of the pages to use it

test/unit/controllers/CommodityCodeBestMatchControllerSpec.scala
```diff
@@ -33,8 +33,9 @@ class CommodityCodeBestMatchControllerSpec extends ControllerSpecBase {
 
   val formProvider = new CommodityCodeBestMatchFormProvider()
   val form: Form[Boolean] = formProvider()
+  val goodsName = "goods"
```
OC-Pomaroff, line 36: We could also change this string to something else and do what @tipperba did here: https://github.com/hmrc/binding-tariff-trader-frontend/pull/204/files#diff-01a5235a65d0e40c32e0876ca64f3d89R57

By using "goods" we aren't actually verifying that the goods name is being read from the cache since that is equal to our default.

PR #214, opened 2020-09-22
DIT-2129 Proposed changes for EORI common component journey

DavidGregory084:


DavidGregory084:
@SteveSugden @PatrykRudnicki could I ask one of you to have a quick look over these changes and make sure we haven't missed anything / that I have the right understanding of the integration between ATaR and ECC?

SteveSugden (commented):
So, to check I have understood...

If your user does not have an HMRC-ATAR-ORG enrolment with an EORINumber identifier, you redirect them to /customs-enrolment-services/atar/subscribe.

If that is your business requirement then that looks good to me.   

However, I was under the impression that you (ATaR team) would be providing the user with a page explaining that they needed to get enrolled with a button or link to /customs-enrolment-services/atar/subscribe

We (ECC) don't have a "start" page explaining to the user what is going on.  That url will land them straight onto our first question page.

DavidGregory084:
Ah cool, thanks @SteveSugden - I will figure out whether we have some work lined up to add that page. As long as we both think that there's nothing extra work needed than checking the enrolments and redirecting then we should be good 👍

PR #217, opened 2020-09-23
Dit 2313

warrenskiwork1979:


test/unit/views/CommodityCodeDigitsViewSpec.scala
```diff
@@ -28,12 +28,14 @@ class CommodityCodeDigitsViewSpec extends StringViewBehaviours {
 
   val form = new CommodityCodeDigitsFormProvider()()
 
-  def createView = () => commodityCodeDigits(frontendAppConfig, form, NormalMode)(fakeRequest, messages)
+  val goodsName = "goods"
```
DavidGregory084, line 31: I think this should have a different name from the default so we can verify we aren't getting the default

test/unit/controllers/CommodityCodeDigitsControllerSpec.scala
```diff
@@ -35,6 +35,7 @@ class CommodityCodeDigitsControllerSpec extends ControllerSpecBase {
 
   val formProvider = new CommodityCodeDigitsFormProvider()
   val form: Form[String] = formProvider()
+  val goodsName = "goods"
```
DavidGregory084, line 38: Same here

DavidGregory084 (approved):
Looks good 👍

test/unit/controllers/CommodityCodeDigitsControllerSpec.scala
warrenskiwork1979: Modified in new commit

conf/messages.en
```diff
@@ -287,9 +287,11 @@ commodityCodeBestMatch.checkYourAnswersLabel = Have you found a commodity code?
 commodityCodeBestMatch.error.required = Select yes if you have found a commodity code
 
 commodityCodeDigits.title = What is the commodity code you found for your item? - Manage your Advance Tariff Rulings - GOV.UK
```
DavidGregory084, line 289: Sorry missed this first time - is the title supposed to be updated too?

PR #221, opened 2020-09-25
DIT 2338 new account dashboard layout implementation

shiromi:


app/views/account_dashboard_statuses.scala.html
```diff
@@ -0,0 +1,127 @@
...
+title = messages("index.title"),
+appConfig = appConfig,
+contentClass = Some("column-full"),
+bodyClasses = None) {
+
+
+<h1 class="heading-xlarge">
+ <span class="heading-secondary">@messages("application.dashboard.status.title")</span>
+ @messages("index.heading")
+</h1>
```
DavidGregory084, line 48: Could we potentially use the `heading` component for this now that it has the `captionMsg` parameter to add a secondary heading?

app/views/account_dashboard_statuses.scala.html
```diff
@@ -0,0 +1,127 @@
...
+    }
+    case CaseStatus.COMPLETED if c.hasExpiredDecision => {
+     <span class="govuk-tag govuk-tag--grey">@messages("case.application.ruling.status.expired")</span>
+    }
+    case CaseStatus.CANCELLED => {
+     <span class="govuk-tag govuk-tag--grey">@messages("case.application.status.cancelled")</span>
+    }
+    case _ => {}
+    }
+   </td>
```
DavidGregory084, line 99: Maybe we should have a helper function or a separate component to figure out the tag to display?

app/views/account_dashboard_statuses.scala.html
```diff
@@ -0,0 +1,127 @@
...
+    </a>
+    <p>
+    @if(Set(CaseStatus.COMPLETED, CaseStatus.CANCELLED).contains(c.status)) {
+    <a id="applications-rulings-list-row-@{index}-view-rulings-link" href="@routes.RulingController.viewRuling(c.reference)">
+     @messages("case.application.ruling.viewRuling")
+    </a>
+    }
+    </p>
+    <p>
+     @if((c.hasActiveDecision && c.daysDifference.get <= 120) | (Set(CaseStatus.COMPLETED).contains(c.status) && c.hasExpiredDecision) | (Set(CaseStatus.CANCELLED).contains(c.status))) {
```
DavidGregory084, line 113: What do you think about adding helper functions for things like `hasExpiringRuling`, `hasActiveRuling` etc. so that we aren't hardcoding numbers like 120 into our views?

app/models/Case.scala
```diff
@@ -46,4 +48,6 @@ case class Case
 
   private val rulingStates = Set(CaseStatus.COMPLETED, CaseStatus.CANCELLED)
 
+  val daysDifference: Option[Long] = this.decision.flatMap(_.effectiveEndDate).map(date => ChronoUnit.DAYS.between(Instant.now, date))
```
DavidGregory084, line 51: I think it might be clearer if this was called `daysUntilExpiry`

app/views/account_dashboard_statuses.scala.html
```diff
@@ -0,0 +1,127 @@
...
+    </p>
+   </td>
+  </tr>
+  }
+  </tbody>
+ </table>
+ @paginate("mt-2", "bottom")
+</div>
+
+}
```
DavidGregory084, line 127: Could we add a `ViewSpec` for this view? That might enable us to up the coverage % back to what it was before

app/config/FrontendAppConfig.scala
```diff
@@ -65,6 +65,8 @@ class FrontendAppConfig @Inject()(
   lazy val digitalcentre = loadConfig("accessibility-urls.digitalcentre")
   //accessibility Urls end
 
+  lazy val accountDashboardUrl = loadConfig("account_dashboard_statuses-urls.account_dashboard_statuses")
```
DavidGregory084, line 68: This config doesn't seem to exist, can you please add it to `application.conf` or remove it if it's not needed?

DavidGregory084 (changes requested):
This looks really good, I've left a couple of comments, let me know what you think!

app/views/account_dashboard_statuses.scala.html
shiromi: Haven't create a VIewSpec as cover it under one of refactoring ticket

DavidGregory084 (approved):
Looks really good 👍 🎉

app/controllers/IndexController.scala
```diff
@@ -75,4 +76,15 @@ class IndexController @Inject()(
...
     }
   }
+  def getApplicationsAndRulings(page: Int): Action[AnyContent] = identify.async { implicit request =>
+    request.eoriNumber match {
+      case Some(eori: String) =>
+        service.getCases(eori, applicationStatuses, SearchPagination(page), Sort()) flatMap { pagedResult =>
+          successful(Ok(account_dashboard_statuses(appConfig, pagedResult)))
+        }
+
+      case None => successful(Redirect(routes.BeforeYouStartController.onPageLoad()))
```
DavidGregory084, line 86: I think that this will need to set the `initialAnswers` like the `getRulings` and `getApplications` functions do

shiromi: I have added the changes. Thanks David

vajarova (approved):

PR #224, opened 2020-09-30
Dit 2307

tipperba:


tipperba:
This is a draft because I'm waiting to hear about one possible change which is in the prototype but not the ticket before merging. It's only a minor change though, so didn't want to continue holding up review.

tipperba:
Sheldon wanted the change mentioned above to be made, so it's been implemented. PR is now open / ready to review.

DavidGregory084:
This is great work @tipperba, but after the navigation changes I we will have to supersede this with a reworked version at #225

PR #225, opened 2020-10-02
DIT-2307 add make file confidential page to the journey

DavidGregory084:
Co-authored-by: David Gregory <david.gregory@opencastsoftware.com>
Co-authored-by: tipperba <66072836+tipperba@users.noreply.github.com>

I reworked these changes in order to try out another idea I had to do with the navigation - still some tidying up to do on Monday such as adding reusable test methods for the `AccumulatingCachingController`s


app/controllers/MakeFileConfidentialController.scala
```diff
@@ -0,0 +1,52 @@
...
+  val getData: DataRetrievalAction,
+  val requireData: DataRequiredAction,
+  formProvider: MakeFileConfidentialFormProvider,
+  cc: MessagesControllerComponents
+)(implicit ec: ExecutionContext) extends MapCachingController[Boolean](cc) {
+  val form = formProvider()
+  val questionPage = MakeFileConfidentialPage
+
+  def renderView(preparedForm: Form[(String, Boolean)], mode: Mode)(implicit request: DataRequest[_]): HtmlFormat.Appendable = {
+    val fileId = request.userAnswers.get(UploadSupportingMaterialMultiplePage).flatMap(_.lastOption).map(_.id).get
```
DavidGregory084, line 49: This is a thing that I preferred in #224 - @tipperba used a parameter to decide which file to delete rather than this - which seems structurally guaranteed to succeed but still makes me uncomfortable due to the `.get`

tipperba (approved):
This looks great. I had a look at it in the context of the earlier navigation PR which is obviously a vast improvement. 

Very much like the way navigation's been restructured and the AnswerCaching / AnswerCachingController type structure looks very well thought out and the code well-expressed. It's very nice to see the way in which all the changes have enabled the controllers to be simplified so dramatically.   

So a big thumbs up.

app/views/components/show_file_table.scala.html
```diff
@@ -15,14 +15,15 @@
...
-@(files: Seq[FileAttachment], mode: Mode)(implicit messages: Messages)
+@(files: Seq[FileView], mode: Mode)(implicit messages: Messages)
 
 <dl class="file-list mb-5">
-    @for((file: FileAttachment, index: Int) <- files.zipWithIndex) {
+    @for((file: FileView, index: Int) <- files.zipWithIndex) {
         <div id="files-row-@{index}" class="file-list-row">
             <dt class="file-list-name" id="files-row-@{index}-title">@file.name</dt>
-            <dd class="file-list-size" id="files-row-@{index}-size" >@humanReadableSize(file.size)</dd>
+            <dd class="file-list-size" id="files-row-@{index}-size" >@confidentialityStatusLabel(file.confidential)</dd>
```
OC-Pomaroff, line 26: Should we update this class and id from `size` to something else more representative?

app/controllers/DeclarationController.scala
```diff
@@ -89,15 +93,20 @@ class DeclarationController @Inject()(
...
+    caseService.create(appendAttachments(newCaseRequest, attachments, confidentialityStatuses, letter, answers))
   }
 
-  private def appendAttachments(caseRequest: NewCaseRequest, attachments: Seq[PublishedFileAttachment],
+  private def appendAttachments(caseRequest: NewCaseRequest,
+                                attachments: Seq[PublishedFileAttachment],
+                                confidentialityStatuses: Map[String, Boolean],
                                 letter: Option[PublishedFileAttachment], answers: UserAnswers): NewCaseRequest = {
-    def toAttachment: PublishedFileAttachment => Attachment = file => Attachment(file.id)
+    def toAttachment: PublishedFileAttachment => Attachment = file => Attachment(file.id, confidentialityStatuses(file.id))
```
OC-Pomaroff, line 109: This map lookup `confidentialityStatuses(file.id)` could throw an exception if the key doesn't exist, should we handle this case?

app/views/addSupportingDocuments.scala.html
```diff
@@ -0,0 +1,60 @@
...
+@import views.ViewUtils._
+
+@(appConfig: FrontendAppConfig, form: Form[_], goodsName: String, mode: Mode)(implicit request: Request[_], messages: Messages)
+
+@main_template(
+    title = s"${errorPrefix(form)} ${messages("addSupportingDocuments.title")}",
+    appConfig = appConfig,
+    bodyClasses = None) {
+
+    @helpers.form(action = AddSupportingDocumentsController.onSubmit(mode)) {
```
OC-Pomaroff, line 30: Is this missing `@CSRF.formField`?

OC-Pomaroff (approved):
Looks like a really neat extension to the navigator refactoring!

app/views/addSupportingDocuments.scala.html
tipperba: I think helpers.form(...) calls CSRF.formField().

OC-Pomaroff: Just tested it and yes you are right - in that case we should possibly remove it from _makeFileConfidential.scala.html_ (and perhaps the others where we have included it).

DavidGregory084: I'm not sure if it does @tipperba looking at https://www.playframework.com/documentation/2.7.x/ScalaCsrf

Thanks @OC-Pomaroff that is a good spot

app/views/makeFileConfidential.scala.html
```diff
@@ -0,0 +1,60 @@
...
+@(appConfig: FrontendAppConfig, makeFileConfidentialForm: Form[_], mode: Mode, fileId: String)(implicit request: Request[_], messages: Messages)
+
+@main_template(
+    title = s"${errorPrefix(makeFileConfidentialForm)} ${messages("makeFileConfidential.title")}",
+    appConfig = appConfig,
+    bodyClasses = None) {
+
+    @helpers.form(action = MakeFileConfidentialController.onSubmit(mode)) {
+
+        @CSRF.formField
```
OC-Pomaroff, line 33: Here - not necessary since it's already included as part of `helpers.form`

app/controllers/DeclarationController.scala
tipperba: This is an interesting one. The structure of the application service flow should always ensure a value is present for the file id, but I guess it's possible it could change in a way which does not mean it's guaranteed.

app/views/addSupportingDocuments.scala.html
tipperba: I think this is the code called (don't shoot the messenger):

`play.filters.csrf.CSRF.getToken(request).map` {token => CSRF.formField(request)}`

app/controllers/DeclarationController.scala
DavidGregory084: Yeah I think this is similar to the use of `last` and `get` in [MakeFileConfidentialController](https://github.com/hmrc/binding-tariff-trader-frontend/pull/225/files#diff-5ae651b06f87aae2d9ee772e62f1ea46). Both of those calls could throw but only if the user has somehow broken the navigation and followed the journey out of order, and I think that is probably worthy of a 500 in any case.

OC-Pomaroff: It's similar, but in this instance it'll throw during submission and we'll potentially have completed some steps (e.g. published the attachments but then fail to create the case).

Would it make sense to validate the two maps first in `onSubmit` and we could potentially combine the validated attachment with the confidentiality boolean to avoid subsequent lookups?

e.g. in `onSubmit` ...

```
    val attachments: Seq[FileAttachment] = answers
      .get(SupportingMaterialFileListPage).map(_.fileAttachments)
      .get(UploadSupportingMaterialMultiplePage)
      .getOrElse(Seq.empty)

    val confidentialityStatuses = answers
      .get(MakeFileConfidentialPage)
      .getOrElse(Map.empty[String, Boolean])

    // validate the attachments and confidentialityStatuses, and combine them somehow?
```

DavidGregory084: Yeah I don't see why not, will look at it this afternoon 👍 

Ideally I don't think we would have this fragmented UserAnswers structure and would rather have an application structure that was correct by construction that we could lens into to update but I think that would be a bigger change than I dare to make 😆

DavidGregory084: FWIW the reason I am a bit cagey about updating the `DeclarationController` is that I know it's going to be deleted in DIT-2339 by @colesadam and @warrenskiwork1979's PR for the check your answers page so I have been trying to change it the smallest amount possible so that I can adapt the changes onto their PR later 😱

OC-Pomaroff: 😱  On the plus side it looks like the code for the onSubmit is moved rather than significantly changed

app/views/addSupportingDocuments.scala.html
DavidGregory084: Turns out the answer is more complicated than this:

* Play has an object called `helper` which has a `form` function that doesn't include the CSRF token
* HMRC has an object called `uk.gov.hmrc.play.views.html.helpers` which has a `form` function which does include the token

Which one was used in a view template seems to have been decided by a coin flip so I've switched them all to the HMRC one.

PR #227, opened 2020-10-06
Dit 2339

warrenskiwork1979:


app/controllers/AddConfidentialInformationController.scala
```diff
@@ -22,56 +22,31 @@ import controllers.actions._
...
-        (value) => {
-          val updatedAnswers = request.userAnswers.set(AddConfidentialInformationPage, value)
-          val answers = if (value) updatedAnswers else updatedAnswers.remove(ProvideConfidentialInformationPage)
-          dataCacheConnector.save(answers.cacheMap).map(
-            _ =>
-              Redirect(navigator.nextPage(AddConfidentialInformationPage, mode)(updatedAnswers))
-          )
-        }
-      )
+  val dataCacheConnector: DataCacheConnector,
```
colesadam, line 36: I think this was a regression in our merge - can we take out the vals again? @warrenskiwork1979

warrenskiwork1979: ha ha I don't think all of our conflict fixes took Adam when looking through the messages file previously, I'll sort this out.

PR #228, opened 2020-10-06
DIT-2238 # implemented welsh language translation.

tipperba:


OC-Pomaroff (approved):
Looks good - my only question would be whether we need to do anything about the back link navigation after you switch the language.

tipperba:
Thanks @OC-Pomaroff - yes, I checked back button behaviour and it seems to work as we would want without me needing to change anything (which I was quite surprised by). I haven't yet looked into why it differs from what I was expecting.

PR #229, opened 2020-10-06
Dit 2339

warrenskiwork1979:


app/controllers/CheckYourAnswersController.scala
```diff
@@ -16,73 +16,144 @@
...
+        ).flatten
       )
     )
 
     Ok(check_your_answers(appConfig, sections))
   }
 
+  def onSubmit(): Action[AnyContent] = (identify andThen getData).async { implicit request: OptionalDataRequest[_] =>
+
+    val answers = request.userAnswers.get // TODO: we should not call `get` on an Option
```
DavidGregory084, line 107: If we add `andThen requireData` at the start of the action then we won't have to do the `.get` above

DavidGregory084 (approved):
Looks good to me 👍

app/controllers/CheckYourAnswersController.scala
```diff
@@ -16,73 +16,144 @@
...
-import service.CountriesService
+import play.api.mvc.{Action, AnyContent, MessagesControllerComponents, Result}
+import service.{CasesService, CountriesService, FileService}
+import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendController
 import utils.CheckYourAnswersHelper
 import viewmodels.AnswerSection
 import views.html.check_your_answers
 
+import scala.concurrent.ExecutionContext.Implicits.global
```
DavidGregory084, line 39: Please can we get rid of this use of the global ExecutionContext and add an `implicit ec` to the controller instead? You can see examples in the other controllers

app/controllers/CheckYourAnswersController.scala
DavidGregory084: Then the type of the `implicit request` above will be `DataRequest[_]` instead

warrenskiwork1979: We've added those changes David

DavidGregory084: Awesome! 👍 Looks good to go

PR #223, opened 2020-09-29
DIT-2349 changes to navigation

DavidGregory084:
This contains the navigation changes from DIT-2349 that Adam and Gorkem have worked on. I think I have updated the rest of the journey and the check your answers navigation.

Most of our controllers now do the same thing. I have captured this in [QuestionCachingController](https://github.com/hmrc/binding-tariff-trader-frontend/compare/DIT-2349?expand=1#diff-9a2789eb46c2be9b0d9810814a465f55).

There are a few major exceptions to this in the journey:

* BeforeYouStartController - this controller creates the initial UserAnswers and sets it into the session
* The file upload / file list pages - they are a bit complex at the moment - I am still considering how to handle that, perhaps by adding an extra page for the first trip through "do you want to add supporting documents"?
* Check your answers / declaration controller - that part is the end of the journey obviously so doesn't fit into the QuestionPage abstraction

PR #231, opened 2020-10-06
Revert "DIT-2349 changes to navigation"

DavidGregory084:
Reverts hmrc/binding-tariff-trader-frontend#223

Reverting so we can merge again without squashing - otherwise fixing conflicts on the other PRs is going to be really difficult


DavidGregory084:
This isn't going to work because Github apparently doesn't sign revert commits from a squash merge

PR #238, opened 2020-10-08
Dit 861

tipperba:


tipperba:
Originally planned to do this using just the HTML 5 History API, but I didn't find a clean solution which worked in all browsers. I'm instead now using JS session storage to maintain a history stack of (different) urls. This works with the language toggle and fixes the reported file upload issue for DIT-861 in the browsers I have tried (Firefox, Chrome, IE11 <in browser stack>).

app/assets/javascripts/bindingtarifftraderfrontend.js
```diff
@@ -128,7 +130,11 @@ $(document).ready(function () {
         e.preventDefault();
         if (window.history && window.history.back && typeof window.history.back === 'function' &&
             (docReferrer !== "" && docReferrer.indexOf(window.location.host) !== -1)) {
-            window.history.back();
+            const historyStack = JSON.parse(sessionStorage.getItem("historyStack")) || [];
+            historyStack.pop();    //take the url for page just exited off the stack
+            const previousUrl = historyStack[historyStack.length - 1]
+            window.location.href = previousUrl
+            sessionStorage.setItem("historyStack", JSON.stringify(historyStack));
```
DavidGregory084, line 137: This looks good but I'd like to see a little more defensive programming here due to issues like this:

Screenshot 2020-10-08 at 16 21 59

DavidGregory084: I realise that we should always get the current page URL on the stack but I just don't trust JS 😆

app/assets/javascripts/bindingtarifftraderfrontend.js
```diff
@@ -139,6 +145,15 @@ $(document).ready(function () {
...
     }
 
+    function updateSessionHistory() {
+        const historyStack = JSON.parse(sessionStorage.getItem("historyStack")) || [];
+        const latestUrl = historyStack[historyStack.length - 1]
+        if (latestUrl !== window.location.href) {
+            historyStack.push(window.location.href)
+            sessionStorage.setItem("historyStack", JSON.stringify(historyStack));
+        }
+    }
```
DavidGregory084, line 155: Might be worth thinking about the same kind of issues here too, although I think the effect will be less problematic here as it will just push the current page URL

DavidGregory084 (approved):

DavidGregory084:
This looks really good to me, but I think the use of sessionStorage is heavily discouraged @tipperba - I think we need to make sure that we are able to use the sessionStorage API in all supported browsers

PR #239, opened 2020-10-08
Dit 2337 - altered applications and ruling page to show different message where there are no applications/rulings

colesadam:


shiromi (approved):
Looks good to me

PR #249, opened 2020-10-19
Dit 2388

tipperba:
Fix for validation issue on supporting material file list page. 

Issue: clicking submit without selecting a radio button remains on the existing page (with a 400), but does not show validation error.


DavidGregory084 (approved):
Thanks! 👍 I guess the controller tests don't check the error rendering, that is surprising!

PR #257, opened 2020-10-27
DIT-2416 | add missing fields to the application complete pdf view

gorkemunsal:
This ticket is to add the missing fields to the application complete pdf view.

Attachment field and Other Rulings fields are added.


DavidGregory084 (approved):
Looks good to me 👍 👍

PR #260, opened 2020-10-28
DIT-2400 | Added Welsh translation for have you found a commodity cod…

warrenskiwork1979:
…e page

### DIFF
--- a/conf/messages.cy
+++ b/conf/messages.cy
@@ -9,7 +9,7 @@ addConfidentialInformation.heading = A ydych am ychwanegu unrhyw wybodaeth gyfri
 addConfidentialInformation.caption = Ynglŷn â’r nwyddau
 addConfidentialInformation.p1 = Ni fydd yr wybodaeth hon yn ymddangos ar wefan y Dyfarniadau Tariffau Uwch.
 addConfidentialInformation.checkYourAnswersLabel = Ychwanegu gwybodaeth gyfrinachol?
-addConfidentialInformation.error.required = Dewiswch ‘Iawn’ os ydych am ychwanegu gwybodaeth gyfrinachol
+addConfidentialInformation.error.required = Dewiswch Iawn os ydych am ychwanegu gwybodaeth gyfrinachol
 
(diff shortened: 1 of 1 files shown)


conf/messages.cy
```diff
@@ -9,4 +9,13 @@ addConfidentialInformation.heading = A ydych am ychwanegu unrhyw wybodaeth gyfri
...
\ No newline at end of file
+addConfidentialInformation.error.required = Dewiswch ‘Iawn’ os ydych am ychwanegu gwybodaeth gyfrinachol
+
+commodityCodeBestMatch.title = A ydych wedi dod o hyd i god nwyddau ar gyfer y nwyddau? – Rheoli’ch Dyfarniadau Tariffau Uwch – GOV.UK
+commodityCodeBestMatch.heading = A ydych wedi dod o hyd i god nwyddau ar gyfer y {0}?
+commodityCodeBestMatch.bodyText = Gall nodi cod nwyddau, hyd yn oed nodi’r 2 rif cyntaf yn unig, gyflymu prosesu’ch cais.
+commodityCodeBestMatch.yesFoundCommodityCode = Iawn
+commodityCodeBestMatch.noHaventFoundCommodityCode = Na
+commodityCodeBestMatch.checkYourAnswersLabel = Wedi dod o hyd i god nwyddau?
+commodityCodeBestMatch.error.required = Dewiswch ‘Iawn’ os ydych wedi dod o hyd i god nwyddau
```
vajarova, line 20: Do we expect the quotes in the UI? I think we would need double quotes in this file for them to look right in the UI? I might be wrong let me know if you've tested that 
Thanks

vajarova: Warren, I spoke with Rachel and she asked us to remove the quotes all together. It doesn't need quotes and is a possible mistake. Could you please do this for the one I can see above under message key addConfidentialInformation.error.required as well 
Thank you

vajarova (approved):

conf/messages.cy
warrenskiwork1979: Have implemented the changes

PR #267, opened 2020-10-30
DIT-2378 | Added toggle for removing the sending sample pages

gorkemunsal:
app config changes for the environments : These should be merged first before merging this PR.

Prod     https://github.com/hmrc/app-config-production/pull/8800
Staging https://github.com/hmrc/app-config-staging/pull/6698
QA:       https://github.com/hmrc/app-config-qa/pull/9207


test/unit/views/ApplicationSubmittedViewSpec.scala
```diff
@@ -192,6 +196,15 @@ class ApplicationSubmittedViewSpec extends ViewBehaviours {
       doc should containText(messages("provideBTIReference.checkYourAnswersLabel"))
     }
 
+    "contain a message to not send a sample when not sending samples toggle is on" in {
```
OC-Pomaroff, line 199:
```suggestion
    "contain a message to not send a sample when sending samples toggle is on" in {
```

OC-Pomaroff (approved):

conf/application.conf
```diff
@@ -214,3 +214,7 @@ accessibility-urls {
 }
 
 host = "http://localhost:9582"
+
+toggle {
+  samples = true
+}
```
DavidGregory084, line 220: Could we review the name of this toggle? 

Should it be called `samplesNotAccepted` or `samplesRejected` or something like that? I think at the moment it's a bit hard to figure out what will happen if this toggle is turned on or off as the default is `true` which seems like it should mean that samples are allowed

gorkemunsal: yeah sounds we have changed it to samplesNotAccepted now

DavidGregory084: Awesome thanks @gorkemunsal

colesadam: I've merged those config PRs after the toggle name change

PR #278, opened 2020-11-06
DIT-1604 Ensure the 'public' property is set correctly for file attachments

DavidGregory084:
This seems to be flipped around, whoops


OC-Pomaroff (approved):
Haha whoops indeed

PR #304, opened 2020-12-15
Dit 2526 - Added new parameter shouldPublishToRuling for the attachment tab in Operational service

emilia-trendafilova:
This should be merged after https://github.com/hmrc/binding-tariff-classification/pull/157


emilia-trendafilova:
After discussion with @OC-Pomaroff  we do not need changes in the trader-frontend. Therefore the PR is closed.

PR #310, opened 2021-01-07
DIT-2451 added try catch block the request is made synchronously

shiromi:


DavidGregory084:
I think this would only be a synchronous request if the `xhr.open` call boolean parameter was `false` but please let me know if I am wrong! (See e.g. [MDN](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open))

PR #340, opened 2021-08-16
DDCE-1978: govuk Assets upgrade and added missing welsh

fraser-sprott:


test/unit/views/behaviours/ViewBehaviours.scala
```diff
@@ -49,13 +48,12 @@ trait ViewBehaviours extends ViewSpecBase {
           val doc = asDocument(view())
           for (key <- expectedGuidanceKeys) assertContainsText(doc, messages(s"$messageKeyPrefix.$key"))
         }
-
         "contain a timeout dialog" in {
-          val doc = asDocument(view())
+          val timeoutElm = asDocument(view()).select("meta[name=\"hmrc-timeout-dialog\"]")
           if(expectTimeoutDialog) {
-            assertRenderedById(doc, "timeoutDialog")
+            timeoutElm.size == 1
```
JMaio, line 54: is this check asserted?

JMaio: does `==` behave in the same way as `shouldBe` or another assertion? it looks like it'll just return `true` or `false` as opposed to failing the test

fraser-sprott: nope, cleary not fully awake yet, will fix it

JMaio (approved):
preliminary approval - would benefit from a second reviewer

yangwong9664 (changes requested):
There are a couple of warnings:

connectors/EmailConnector.scala:37:80: parameter value rds in method send is never used
[warn]   def send[E >: Email[_]](e: E)(implicit hc: HeaderCarrier, writes: Writes[E], rds: Reads[E]): 
                                                                          
utils/JsonFormatters.scala:50:68: pattern var obj in value $anonfun is never used; `obj@_' suppresses this warning

                                                                 ^
connectors/EmailConnector.scala:40:18: value readRaw in object HttpReads is deprecated (since 11.0.0): Use uk.gov.hmrc.http.HttpReads.Implicits instead. See README for differences.
[warn]       client.POST(url = url, body = e).map(_ => ())

There's also a few Scalastyle warnings

yangwong9664 (approved):

JMaio (approved):

PR #355, opened 2022-09-13
DDCE-3376 Some dependency upgrades + removed all Scalastyle warnings

Taiwo-AB:


project/AppDependencies.scala
```diff
@@ -5,7 +5,7 @@ object AppDependencies {
   import play.core.PlayVersion
 
   private val silencerVersion      = "1.7.9"
-  private val bootstrapPlayVersion = "7.2.0"
+  private val bootstrapPlayVersion = "7.3.0"
   private val hmrcMongoPlayVersion = "0.71.0"
```
frikit, line 9: Can be -> uk.gov.hmrc.mongo:hmrc-mongo-play-28           : 0.71.0  -> 0.72.0

frikit (changes requested):
- Please remove integration test settings and mentions as it is not used in the project build.sbt:15 , build.sbt:3
- Please update as well in README.MD:
from: sbt test it:test
to: sbt test

project/AppDependencies.scala
Taiwo-AB: Yh just seen that there is a new release. It was 0.71.0 when i checked. Will bump :)

PR #374, opened 2023-04-27
DDCE-4185 - Update CSP and fix duplicate ruling cases on check your an…

mywyau:
Summary: 

Updated the CSP and fix the issue with the ATs. There is a duplicating similar rulings bug in main. This has to do with the caching mechanism in the trader frontend. The proposed solution is to clear down the cache to stop it before adding the new page values to the cache map. This way when you submit on the page before the Check Answer page it does not duplicate the previous set of data and only adds the new value to the list of values.

E.g.  It basically keeps doubling and and stores the values and can reach 10,000s of ruling references. The live in mongo time for this cache is 1 hour in app.config. If running the Traders tests a lot locally this can get out of hand.

Screenshot 2023-04-27 at 10 38 43

Screenshot 2023-04-27 at 10 38 57

Screenshot 2023-04-27 at 10 39 22

Screenshot 2023-04-27 at 10 39 29


EdCope (changes requested):
Please remove the original filter in application.conf

EdCope:
Adding comment to remind about js-visible issue on supportingMaterialFileList not showing files due to CSP

PR #375, opened 2023-04-27
DDCE-4144 - remove /unauthorised route from project as it is not used

frikit:


conf/messages.cy
```diff
@@ -69,8 +69,6 @@ site.cancel_application.hiddenText=a mynd yn ôl i dudalen eich ceisiadau a dyfa
 site.sectionIs = Yr adran hon yw
 site.save_your_ruling_certificate=Cadw’r dystysgrif dyfarniad ar ffurf PDF
 site.print_your_ruling_certificate=Argraffu’ch tystysgrif dyfarniad
-site.save_your_cover_letter=Cadw’r llythyr eglurhaol ar ffurf PDF
```
frikit, line 72: this was duplicate, look above

app/controllers/actions/IdentifierAction.scala
```diff
@@ -69,19 +68,13 @@ class AuthenticatedIdentifierAction @Inject() (
           redirectToLogin
         case _: InsufficientEnrolments =>
           redirectToEoriComponent
-        case _: InsufficientConfidenceLevel | _: UnsupportedAuthProvider | _: UnsupportedAffinityGroup |
```
frikit, line 72: you cant reach this branch from user journey perspective as we dont request that data from user so this branch cant be triggered

EdCope (changes requested):
Optimize import in AuthActionSpec
and rename test https://github.com/hmrc/binding-tariff-trader-frontend/blob/f56cdddafb4c46c10b19d8fbc13fe4b728c20bed/test/controllers/actions/AuthActionSpec.scala#L78 I think wasn't named right before

EdCope (approved):

PR #383, opened 2023-08-17
DDCE-2412 Fix styling in print window

mywyau:
Spike write up
https://confluence.tools.tax.service.gov.uk/display/ELSE/DDCE-2412+-+ATaR%3A+Investigate+Text+Splitting+in+Issued+Certificates

1. no tests yet just a branch to mess around in. 
2. Re-design of the ruling cert and ruling cover letter
3. Needs to move logic our of view and probably move text splitting into a helper or service and improve the logic
4. css styling can be cleaned up and remove all unused for each custom view. 
5. There is shared styling which I think needs revising as well. Good to be able to split out each letter and not share anything as it makes it hard to change if multiple views use the same custom styling.
6. Finally strip out the pdf generator as pdfs in general are heavily discouraged, the pdf-generator-service is deprecated and  this is petty but it looks ugly.


mywyau:
Screenshot 2023-08-30 at 11 58 54

Screenshot 2023-08-30 at 11 59 04

Screenshot 2023-08-30 at 11 58 59

Screenshot 2023-08-30 at 11 59 09

Screenshot 2023-08-30 at 11 59 15

PR #385, opened 2023-09-11
DDCE-2593 - delete unused urls

frikit:


project/AppDependencies.scala
```diff
@@ -21,8 +20,8 @@ object AppDependencies {
     "com.github.tomakehurst" % "wiremock-standalone"      % "2.27.2",
```
frikit, line 20: cant update this one compplains on missing logback impl

frikit: fixed

lmugnai:
Can we update these?:
org.wiremock:wiremock-standalone:test : 3.0.4                     -> 3.1.0                            
uk.gov.hmrc:play-frontend-hmrc        : 7.19.0-play-28            -> 7.20.0-play-28

lmugnai:
There are some unused vals in FrontendAppConfig that can be deleted.

PR #413, opened 2024-09-16
DDCE-4966: ATaR- Improve the error logging format

Divya-Manivannan-hmrc:


lmugnai:
@Divya-Manivannan-hmrc the formatting changes look a bit strange. Can you make sure you're using the scalafmt.conf file when formatting. I think we need check some of those changes and revert them.

lmugnai (approved):

PR #431, opened 2025-04-14
DDCE-6578 Improve confidentialLabel error handling and logging

lmugnai:


app/utils/CheckYourAnswersHelper.scala
```diff
@@ -168,7 +170,15 @@ class CheckYourAnswersHelper(
       .getOrElse(Map.empty)
 
     def confidentialLabel(attachment: FileAttachment)(implicit messages: Messages): String =
-      if (keepConfidential(attachment.id)) "- " + messages("site.keep_confidential") else ""
+      keepConfidential.get(attachment.id) match {
+        case Some(true) => "- " + messages("site.keep_confidential")
+        case None =>
+          logger.error(s"No confidentiality setting for attachment ID: ${attachment.id}")
```
Divya-Manivannan-hmrc, line 176: Do we need both logger.error and the exception? Can't we just go with the exception as that will also trigger an alert?

Divya-Manivannan-hmrc (changes requested):

Divya-Manivannan-hmrc (approved):

PR #434, opened 2025-05-20
DDCE-6601: Ensure File Name Persistence After Re-uploading Documents on /upload-documents

SachaMcCormack:
1. uncommented code 
2. added javascript to reset file picker as the file does not persist once removed

### DIFF
--- a/app/controllers/MakeFileConfidentialController.scala
+++ b/app/controllers/MakeFileConfidentialController.scala
@@ -54,17 +54,15 @@ class MakeFileConfidentialController @Inject() (
     request: DataRequest[?]
   ): HtmlFormat.Appendable = {
     val fileId: String = request.userAnswers.get(UploadSupportingMaterialMultiplePage).map(_.last.id).get
-      makeFileConfidentialView(appConfig, preparedForm, submitAction, mode, fileId)
+    makeFileConfidentialView(appConfig, preparedForm, submitAction, mode, fileId)
   }
 
   override def onPageLoad(mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData) {
-    implicit request:DataRequest[?] =>
-      val files = request.userAnswers.get(UploadSupportingMaterialMultiplePage)
-      if(files.get.isEmpty){
-        val fileId = request.id.toString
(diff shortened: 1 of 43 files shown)


app/views/uploadSupportingMaterialMultiple.scala.html
```diff
@@ -83,6 +83,11 @@
             document
                 .getElementById("file")
                 .addEventListener("change", onFileSelected(id, targetURL, csrfToken));
+
+            window.addEventListener("pageshow", function(event) {
+                const files = document.querySelectorAll('input[type="file"]');
+                files.forEach(input => input.value = '');
+            });
```
lmugnai, line 90: We shouldn't be adding nested javascript like this. If needed keep it in a separate js file. However, we are encouraged to not use javascript at all on our services. Can we look into an alternative solution?

app/controllers/MakeFileConfidentialController.scala
```diff
@@ -54,7 +54,18 @@ class MakeFileConfidentialController @Inject() (
     request: DataRequest[?]
   ): HtmlFormat.Appendable = {
     val fileId: String = request.userAnswers.get(UploadSupportingMaterialMultiplePage).map(_.last.id).get
-    makeFileConfidentialView(appConfig, preparedForm, submitAction, mode, fileId)
+      makeFileConfidentialView(appConfig, preparedForm, submitAction, mode, fileId)
+  }
+
+  override def onPageLoad(mode: Mode): Action[AnyContent] = (identify andThen getData andThen requireData) {
+    implicit request:DataRequest[?] =>
+      val files = request.userAnswers.get(UploadSupportingMaterialMultiplePage)
```
lmugnai, line 62: if files is None, calling files.get will throw a NoSuchElementException at runtime, causing a server error. Using the followng pattern matching should guard against this too

      request.userAnswers.get(UploadSupportingMaterialMultiplePage) match {
        case Some(files) if files.nonEmpty => Ok(renderView(form, submitAction(mode), mode))
        case _ => Redirect(routes.UploadSupportingMaterialMultipleController.onPageLoad(Some(request.id.toString), mode))
      }

lmugnai (dismissed):

gastonracui (dismissed):

PR #450, opened 2026-06-02
DDCE-8555: Use hmrcStandardPage + dep updates

lmugnai:
Refactored main layout wrapper to use HmrcStandardPage instead of GovukLayout.


Shamanth9798 (approved):
LGTM

